PR #4, opened 2025-12-27
Feature/pluggable converters

felipeblazing:
This PR is meant to add a converters that can be registered and invoked in order to convert between two different data representations that might be on different memory tiers. 

We added a registry where users can register converters. 
The building blocks for invoking these conversions from idata_representations
Implementations of two converters to/from CPU  fixed size buffer reprsentation to cudf::Table

moved some implementations out of headers, consolidated testing code, and fixed some issues with copyright notices

### DIFF
--- a/src/data/representation_converter.cpp
+++ b/src/data/representation_converter.cpp
@@ -22,6 +22,7 @@
 #include "data/gpu_data_representation.hpp"
 #include "memory/host_table.hpp"
 
+#include <rmm/detail/error.hpp>
 #include <rmm/device_buffer.hpp>
 #include <rmm/device_uvector.hpp>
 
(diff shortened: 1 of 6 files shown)


src/data/representation_converter.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+  auto it = _converters.find(key);
+  if (it == _converters.end()) {
+    std::ostringstream oss;
+    oss << "No converter registered for source type '" << key.source_type.name()
+        << "' to target type '" << key.target_type.name() << "'";
+    throw std::runtime_error(oss.str());
+  }
+
+  return it->second(source, target_memory_space, stream);
```
bwyogatama: It does seem like right now the mutex is hold during the conversion? Maybe we should release the mutex during as we only need it to look up the registry?

include/data/representation_converter.hpp
```diff
@@ -0,0 +1,257 @@
...
+   * @example
+   * auto result = registry.convert<gpu_table_representation>(source, space);
+   */
+  template <typename TargetType>
+  std::unique_ptr<TargetType> convert(idata_representation& source,
+                                      const memory::memory_space* target_memory_space,
+                                      rmm::cuda_stream_view stream = rmm::cuda_stream_default) const
+  {
+    converter_key key{std::type_index(typeid(source)), std::type_index(typeid(TargetType))};
+    auto result = convert_impl(key, source, target_memory_space, stream);
```
bwyogatama: Maybe there should be a check in case the converter is found but the the target memory space is invalid?

For example if someone converting gpu_data_representation to host_data_representation but targeting another GPU memory space, it should be invalid

include/data/representation_converter.hpp
```diff
@@ -0,0 +1,257 @@
...
+
+/**
+ * @brief Initialize the built-in representation converters.
+ *
+ * This function registers the default converters between gpu_table_representation
+ * and host_table_representation.
+ *
+ * @param registry The converter registry to register converters with.
+ */
+void register_builtin_converters(representation_converter_registry& registry);
```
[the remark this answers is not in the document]
felipeblazing: One way to manage this might be to expose an API in libcudf that allows you to get back the metadata like in pack but with a vector of rmm::device_buffers that contain all of the relevant data.

PR #95, opened 2026-03-30
Fast path for same-device GPU-to-GPU conversion

mbrobbel:
Skip cudf::pack/unpack round-trip when source and target are on the same GPU device — clone directly via the source representation instead. I was hitting the `assert`.


felipeblazing:
This makes sense. In the future we should copy withou thte pack and unpack the same way we do for host tables

felipeblazing (approved):

PR #96, opened 2026-04-02
Feature/file downgrade

felipeblazing:
Add the ability to downgrade files. We are still experimenting with different approaches so this isn't ready to be merged in


benchmark/benchmark_main.cpp
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <benchmark/benchmark.h>
+
+BENCHMARK_MAIN();
```
joosthooz: Do we really need to separate this out into a new file?

src/data/kvikio_io_backend.cpp
```diff
@@ -0,0 +1,148 @@
...
+#include <cstring>
+#include <future>
+#include <string>
+
+namespace cucascade {
+namespace {
+
+/// Number of threads for kvikIO parallel I/O (pread/pwrite).
+/// Each thread submits an independent I/O request, saturating NVMe queue depth.
+constexpr unsigned int KVIKIO_NUM_THREADS = 16;
```
joosthooz: Why hardcode this instead of allowing users to set the KVIKIO env vars how they see fit?

src/data/pipeline_io_backend.cpp
```diff
@@ -0,0 +1,382 @@
...
+constexpr std::size_t DIRECT_IO_ALIGNMENT = 512;
+
+/// Round up to O_DIRECT alignment boundary
+std::size_t align_up_dio(std::size_t n)
+{
+  return (n + DIRECT_IO_ALIGNMENT - 1) & ~(DIRECT_IO_ALIGNMENT - 1);
+}
+
+// =============================================================================
+// pipeline_io_backend — double-buffered pinned host memory pipeline
```
joosthooz: Is it an idea to use more than 2 buffers to drive more parallel I/O operations? Or if we already issue multiple parallel I/O operations into the currently active buffers, that's ok too.

src/data/pipeline_io_backend.cpp
```diff
@@ -0,0 +1,382 @@
...
+      std::size_t next_chunk = 0;
+      if (remaining > 0) {
+        next_chunk     = std::min(remaining, PIPELINE_BUF_SIZE);
+        auto read_sz   = _direct_io ? align_up_dio(next_chunk) : next_chunk;
+        int other      = 1 - cur;
+        auto* buf_ptr  = _buf[other];
+        auto read_off  = src_offset;
+        auto actual_sz = next_chunk;
+        auto read_fd   = fd;
+        read_future    = std::async(std::launch::async,
```
joosthooz: This creates a new thread for every chunk, we may want to consider using a threadpool instead. Not just for the overhead, but to have a handle on how many of these I/O operations are active at the same time

joosthooz (commented):
I didn't see everything yet, but wanted to post some initial questions

joosthooz:
Other question: can't we use the (pretty new) cudftable format for this purpose? https://docs.rapids.ai/api/libcudf/stable/group__io__writers#ga18b2cd6a0deed1c4fbc43c2fc6b3094d

felipeblazing:
> Other question: can't we use the (pretty new) cudftable format for this purpose? https://docs.rapids.ai/api/libcudf/stable/group__io__writers#ga18b2cd6a0deed1c4fbc43c2fc6b3094d

Using pack means you need more gpu memory to downgrade when you are already strapped for memory. This avoids that. We could pr this to cudf at some point.

include/cucascade/data/disk_io_backend.hpp
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * @brief Read data from a disk file into device memory.
+   *
+   * @param path File path to read from.
+   * @param dev_ptr Pointer to device memory to read into.
+   * @param size Number of bytes to read.
+   * @param file_offset Byte offset within the file to start reading from.
+   * @param stream CUDA stream for synchronization.
+   */
+  virtual void read_device(const std::string& path,
```
aminaramoon: take filesystem path instead of string

include/cucascade/data/disk_io_backend.hpp
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * @brief Write data from device memory to a disk file.
+   *
+   * @param path File path to write to.
+   * @param dev_ptr Pointer to device memory containing the data.
+   * @param size Number of bytes to write.
+   * @param file_offset Byte offset within the file to start writing at.
+   * @param stream CUDA stream for synchronization.
+   */
+  virtual void write_device(const std::string& path,
```
aminaramoon: you could take in a cudf::device_span, then you don't need to create write_device, write_host and ...

you could just add write and read

include/cucascade/data/disk_file_format.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * @brief Align a value up to the specified power-of-two alignment.
+ *
+ * @param value The value to align.
+ * @param alignment The alignment boundary (must be a power of two).
+ * @return The smallest multiple of alignment >= value.
+ */
+[[nodiscard]] constexpr std::size_t align_up(std::size_t value, std::size_t alignment) noexcept
+{
```
aminaramoon: since rmm is a dependecy, it already proivide align_up and down for us

aminaramoon (changes requested):

src/data/representation_converter.cpp
```diff
@@ -819,6 +912,644 @@ std::unique_ptr<idata_representation> convert_host_fast_to_host_fast(
...
+  if (header.metadata_size > 0) {
+    backend.read_host(
+      file_path, metadata_bytes.data(), header.metadata_size, sizeof(disk_file_header));
+  }
+
+  auto disk_columns = deserialize_column_metadata(metadata_bytes.data(), metadata_bytes.size());
+
+  // Set CUDA device to target GPU
+  int previous_device = -1;
+  CUCASCADE_CUDA_TRY(cudaGetDevice(&previous_device));
```
aminaramoon: rmm has set_device_id_raii that does this, should be better to use that isntead

felipeblazing:
/ok to test [6db1002](https://github.com/NVIDIA/cuCascade/pull/96/commits/6db1002b8a4bbd60b15e98f166070045390d7118)

benchmark/benchmark_main.cpp
aminaramoon: you can remove this file, 
just link against benchmakr_main and it provides the main for you @felipeblazing

src/data/disk_data_representation.cpp
```diff
@@ -0,0 +1,84 @@
...
+// =============================================================================
+
+namespace memory {
+
+namespace {
+std::atomic<uint64_t> g_disk_file_counter{0};
+}  // namespace
+
+std::string generate_disk_file_path(std::string_view base_path)
+{
```
aminaramoon, line 75: you will (unlikel) have name collision when let's in doris mathias is running two sirius in the same machine.
instead use 

```
#include <cstdlib>
#include <unistd.h>

std::filesystem::path create_secure_temp_file(std::string base_path) {
    std::string pattern = base_path + "/batch_XXXXXX.cucascade";
    int fd = mkstemps(pattern.data(), 10); // 10 is the length of ".cucascade"
    if (fd != -1) {
        close(fd); // You now have a guaranteed unique file created
        return pattern;
    }
    return {};
}
```

that guarantees a new file that no one else can claim

aminaramoon (commented):
left some comments

aminaramoon (approved):

felipeblazing:
/ok to test [0f6d2ad](https://github.com/NVIDIA/cuCascade/pull/96/commits/0f6d2ad3e0cc42266c672a32f3aa76ba358342a5)

PR #99, opened 2026-04-06
feat\!: Replace data_batch state machine with RAII lock accessor types

bwyogatama:
## Summary

Replace the 4-state finite state machine (`batch_state`: idle, task_created, processing, in_transit) with a 3-class design where `data_batch` is the "idle" state and all data access requires acquiring a lock through RAII accessor types.

**Core invariant:** it is impossible to read or mutate batch data without holding the appropriate lock, and move semantics make stale references a compile error.

## Design

```
data_batch (idle — owns shared_mutex + data representation)
├── Lock-free public API: get_batch_id(), subscribe(), unsubscribe(), get_state()
├── Private data accessors: get_data(), get_current_tier(), set_data()
│   └── Only accessible by friend accessor classes
│
├── read_only_data_batch (RAII shared lock, holds shared_ptr<data_batch>)
│   └── Read accessors + clone() + clone_to()
│
└── mutable_data_batch (RAII exclusive lock, holds shared_ptr<data_batch>)
    └── Read accessors + clone() + clone_to() + set_data() + convert_to()
```

### Observable state

`data_batch` exposes an atomic `batch_state` enum (`idle`, `read_only`, `mutable_locked`) via `get_state()`. Updated during every state transition. Enables the repository to filter batches by lock state.

### State transitions

**Non-static methods** (use `shared_from_this` — caller's `shared_ptr` is NOT consumed):
```cpp
auto accessor = batch->to_read_only();   // batch still valid, state = read_only
auto accessor = batch->to_mutable();     // batch still valid, state = mutable_locked
```

**Static methods** (consume accessors via move):
```cpp
auto batch = data_batch::to_idle(std::move(ro));         // ro is now consumed
```

**Locked-to-locked transitions** (static, consume via move):
```cpp
auto mut = data_batch::readonly_to_mutable(std::move(ro));   // upgrade lock
auto ro = data_batch::mutable_to_readonly(std::move(mut));   // downgrade lock
```

Non-blocking variants (`try_to_read_only`, `try_to_mutable`) return `std::optional`.

### State-filtered repository

`pop_data_batch()` replaced with state-specific FIFO pops that filter by `get_state()`:
```cpp
auto idle_batch = repo.pop_idle_data_batch();       // first batch where state == idle
auto ro_batch   = repo.pop_read_only_data_batch();  // first batch where state == read_only
auto mut_batch  = repo.pop_mutable_data_batch();    // first batch where state == mutable_locked
```

`pop_data_batch_by_id()` and `get_data_batch_by_id()` unchanged — caller checks state after retrieval.

## Breaking changes

- `data_batch` now inherits `std::enable_shared_from_this<data_batch>`
- `data_batch` has observable `batch_state` enum and `get_state()` method
- `batch_state` enum replaced (old: idle/task_created/processing/in_transit → new: idle/read_only/mutable_locked)
- **`read_only_data_batch` and `mutable_data_batch` are no longer class templates** — they always hold `std::shared_ptr<data_batch>` (the `unique_ptr` variant was unused)
- Static entry-point methods removed (`to_read_only(PtrType&&)`, `to_mutable(PtrType&&)`, `try_to_read_only(PtrType&)`, `try_to_mutable(PtrType&)`) — use non-static `shared_from_this` variants instead
- `convert_to` moved from `data_batch` (private) to `mutable_data_batch` (public)
- `mutable_data_batch` now has `clone()` and `clone_to()` (matching `read_only_data_batch`)
- `pop_data_batch()` removed — use `pop_idle_data_batch()`, `pop_read_only_data_batch()`, or `pop_mutable_data_batch()`
- `idata_batch_probe`, `data_batch_processing_handle`, `lock_for_processing_result/status` removed
- `pop_data_batch_by_id()` and `get_data_batch_by_id()` no longer take `target_state`
- `data_repository_manager::add_data_batch_impl` uses `if constexpr` instead of SFINAE

## Files changed

| Area | Files | What |
|------|-------|------|
| Core type system | `data_batch.hpp`, `data_batch.cpp` | 3-class rewrite, observable state, `enable_shared_from_this`, locked↔locked transitions, de-templatized accessors |
| Repository layer | `data_repository.hpp` | State-filtered `pop_idle/read_only/mutable_data_batch()` |
| Repository impl | `data_repository.cpp`, `data_repository_manager.cpp` | Updated type references |
| Converter | `representation_converter.cpp` | Added new batch type support |
| Tests | `test_data_batch.cpp`, `test_data_repository.cpp`, `test_data_repository_manager.cpp` | Full rewrite + new state/transition/filtered-pop tests |

## Test plan

- [x] `pixi run build` compiles cleanly (all 68 targets)
- [x] `pixi run test` passes (43 test cases, 1116 assertions for `[data_batch]`)
- [x] Non-blocking `try_to_read_only` / `try_to_mutable` return `nullopt` when lock is held
- [x] `get_state()` returns correct value after every transition type
- [x] State-filtered pops skip batches in wrong state
- [x] Non-static transitions preserve caller's `shared_ptr`
- [x] `readonly_to_mutable` / `mutable_to_readonly` upgrade/downgrade correctly
- [x] `clone()` / `clone_to()` work on both `read_only_data_batch` and `mutable_data_batch`
- [x] `convert_to` on `mutable_data_batch` works with GPU sync
- [x] Accessor classes compile without template parameter (de-templatized)

### DIFF
--- a/include/cucascade/data/data_batch.hpp
+++ b/include/cucascade/data/data_batch.hpp
@@ -39,6 +39,15 @@ class memory_space;
 
 namespace cucascade {
 
+/**
+ * @brief Observable state of a data_batch.
+ *
+ * Tracks whether the batch is idle, shared-locked (read_only), or
+ * exclusively-locked (mutable_locked). Updated atomically during
+ * state transitions.
+ */
+enum class batch_state { idle, read_only, mutable_locked };
+
 // Forward declarations -- required before data_batch because it friends them.
 template <typename PtrType>
 class read_only_data_batch;
@@ -60,7 +69,7 @@ class mutable_data_batch;
  * @note Non-copyable and non-movable. The object itself never moves; only the
  *       smart pointer to it is transferred between states.
  */
-class data_batch {
+class data_batch : public std::enable_shared_from_this<data_batch> {
  public:
   // -- Construction --
 
@@ -119,8 +128,17 @@ class data_batch {
    */
   size_t get_subscriber_count() const;
 
+  /**
+   * @brief Get the observable lock state of this batch.
+   *
+   * Atomic, lock-free. Returns the current state: idle, read_only, or
+   * mutable_locked. Updated during every state transition.
+   *
+   * @return The current batch_state.
+   */
+  batch_state get_state() const { return _state.load(std::memory_order_relaxed); }
+
   // -- Static transition methods (D-13/D-14/D-15/D-17) --
-  // All transitions go through idle -- no locked-to-locked shortcuts.
 
   /**
    * @brief Transition from idle to read-only (shared lock).
@@ -197,6 +215,76 @@ class data_batch {
   template <typename PtrType>
   [[nodiscard]] static std::optional<mutable_data_batch<PtrType>> try_to_mutable(PtrType& batch);
 
+  // -- Non-static transitions (shared_ptr only, via shared_from_this) --
+  // The caller's shared_ptr is NOT consumed. These only work when the
+  // data_batch is managed by a shared_ptr (throws bad_weak_ptr otherwise).
+
+  /**
+   * @brief Transition from idle to read-only (shared lock) without consuming the caller's pointer.
+   *
+   * Uses shared_from_this() to obtain a new shared_ptr. Blocks until the
+   * shared lock is acquired.
+   *
+   * @return A read_only_data_batch holding the shared lock.
+   */
+  [[nodiscard]] read_only_data_batch<std::shared_ptr<data_batch>> to_read_only();
+
+  /**
+   * @brief Transition from idle to mutable (exclusive lock) without consuming the caller's pointer.
+   *
+   * Uses shared_from_this() to obtain a new shared_ptr. Blocks until the
+   * exclusive lock is acquired.
+   *
+   * @return A mutable_data_batch holding the exclusive lock.
+   */
+  [[nodiscard]] mutable_data_batch<std::shared_ptr<data_batch>> to_mutable();
+
+  /**
+   * @brief Try to transition from idle to read-only (non-blocking, shared_ptr only).
+   *
+   * @return An optional containing the read-only accessor on success, or
+   *         std::nullopt if the lock could not be acquired immediately.
+   */
+  [[nodiscard]] std::optional<read_only_data_batch<std::shared_ptr<data_batch>>> try_to_read_only();
+
+  /**
+   * @brief Try to transition from idle to mutable (non-blocking, shared_ptr only).
+   *
+   * @return An optional containing the mutable accessor on success, or
+   *         std::nullopt if the lock could not be acquired immediately.
+   */
+  [[nodiscard]] std::optional<mutable_data_batch<std::shared_ptr<data_batch>>> try_to_mutable();
+
+  // -- Locked-to-locked static transitions --
+
+  /**
+   * @brief Transition from read-only to mutable (upgrade lock).
+   *
+   * Releases the shared lock, then acquires an exclusive lock (may block).
+   * The source accessor is consumed via move.
+   *
+   * @tparam PtrType Smart pointer type (shared_ptr or unique_ptr to data_batch).
+   * @param accessor Rvalue reference to the read-only accessor (consumed).
+   * @return A mutable_data_batch holding the exclusive lock.
+   */
+  template <typename PtrType>
+  [[nodiscard]] static mutable_data_batch<PtrType> readonly_to_mutable(
+    read_only_data_batch<PtrType>&& accessor);
+
+  /**
+   * @brief Transition from mutable to read-only (downgrade lock).
+   *
+   * Releases the exclusive lock, then acquires a shared lock (may block).
+   * The source accessor is consumed via move.
+   *
+   * @tparam PtrType Smart pointer type (shared_ptr or unique_ptr to data_batch).
+   * @param accessor Rvalue reference to the mutable accessor (consumed).
+   * @return A read_only_data_batch holding the shared lock.
+   */
+  template <typename PtrType>
+  [[nodiscard]] static read_only_data_batch<PtrType> mutable_to_readonly(
+    mutable_data_batch<PtrType>&& accessor);
+
  private:
   // -- Friend declarations (D-24/REPO-04) --
   template <typename PtrType>
@@ -245,10 +333,11 @@ class data_batch {
                   rmm::cuda_stream_view stream);
 
   // -- Members --
-  const uint64_t _batch_id;                     ///< Immutable batch identifier
-  std::unique_ptr<idata_representation> _data;  ///< Owned data representation
-  mutable std::shared_mutex _rw_mutex;          ///< Reader-writer mutex
-  std::atomic<size_t> _subscriber_count{0};     ///< Atomic subscriber interest count
+  const uint64_t _batch_id;                            ///< Immutable batch identifier
+  std::unique_ptr<idata_representation> _data;         ///< Owned data representation
+  mutable std::shared_mutex _rw_mutex;                 ///< Reader-writer mutex
+  std::atomic<size_t> _subscriber_count{0};            ///< Atomic subscriber interest count
+  std::atomic<batch_state> _state{batch_state::idle};  ///< Observable lock state
 };
 
 /**
@@ -454,6 +543,7 @@ read_only_data_batch<PtrType> data_batch::to_read_only(PtrType&& batch)
 {
   auto ptr = std::move(batch);
   std::shared_lock<std::shared_mutex> lock(ptr->_rw_mutex);
+  ptr->_state.store(batch_state::read_only, std::memory_order_relaxed);
   return read_only_data_batch<PtrType>(std::move(ptr), std::move(lock));
 }
 
@@ -464,6 +554,7 @@ mutable_data_batch<PtrType> data_batch::to_mutable(PtrType&& batch)
 {
   auto ptr = std::move(batch);
   std::unique_lock<std::shared_mutex> lock(ptr->_rw_mutex);
+  ptr->_state.store(batch_state::mutable_locked, std::memory_order_relaxed);
   return mutable_data_batch<PtrType>(std::move(ptr), std::move(lock));
 }
 
@@ -473,6 +564,7 @@ template <typename PtrType>
 PtrType data_batch::to_idle(read_only_data_batch<PtrType>&& accessor)
 {
   auto ptr = std::move(accessor._batch);
+  ptr->_state.store(batch_state::idle, std::memory_order_relaxed);
   accessor._lock.unlock();
   return ptr;
 }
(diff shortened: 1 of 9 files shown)


bwyogatama:
One thing that I still don't know is the best API to pop `data_batch` from the repository with this new model. Previously we passed in the `batch_state` that we want to transition it into, which I think is not a good API but it allows us to skip `data_batch` that does not meet the state machine requirement to transition to that state. 

For example, when popping data batch from the repo for downgrade it allows us to skip data batch that's currently in the `in_processing`. 

But right now, since the state is gone, we will just pop `data_batch` from repository in a FIFO manner and it's up to the caller what do they want to do with it. The problem is now we might run into a 'pop and then insert again' scenarios where we pop a data batch, notice that we cannot downgrade cause a task is executing on it, and have to insert again to the repo.

Let me know if people have some ideas @aminaramoon @dhruv9vats

dhruv9vats:
Thanks for the refactor @bwyogatama !

> I think we cannot do enable_shared_from_this because we want data_batch to work for shared_ptr<data_batch> and unique_ptr<data_batch>. Sirius use shared_ptr<data_batch> but other engine might want to use unique_ptr<data_batch>

These are, IMO, 2 different use cases. The current use case we are addressing is: we want to be able to have multiple readers + enforce a single writer in a streamlined fashion using standard sync utilities (shared_lock / unique_lock). If we remove `shared_from_this`, we lose the ability to extend the lifetime of the `synchronized_data_batch` when we create a read_only (shallow) copy of the object. Which means, the responsibility of managing the lifetime of the original object now again lies in the hands of the end user, since we are propagating raw pointers when we get the read-only / mutable _views_ of the `synchronized_data_batch`. I believe we dont want to do that especially since we have the option to ensure it ourselves.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
+ * synchronized_data_batch owns an inner data_batch (pure data, no locking) and a
+ * std::shared_mutex. Access to the inner data_batch is only possible through RAII accessor
+ * types (read_only_data_batch, mutable_data_batch) that bundle lock lifetime with data access.
  *
- * Applications may implement this interface to hold additional application specific
- * data_batch metadata while probing the data_batch by overriding the provided methods
- * that expose the data_batch state when certain events occur, like state transitions.
+ * Key characteristics:
+ * - Compiler-enforced read/write separation via accessor types
+ * - PtrType agnostic — works with shared_ptr, unique_ptr, or stack allocation
```
dhruv9vats: I think we dont want this to be PtrType agnostic, we are intentionally keeping it so this is only shared_ptr.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
-   * @brief Construct a handle from a shared_ptr (stores weak_ptr; does not keep batch alive).
-   *
-   * @param batch shared_ptr to the data_batch to manage
-   */
-  explicit data_batch_processing_handle(std::shared_ptr<data_batch> batch)
-    : _batch(batch ? std::optional<std::weak_ptr<data_batch>>(std::move(batch)) : std::nullopt)
-  {
-  }
+   private:
+    friend class synchronized_data_batch;
```
dhruv9vats: It is okay to be explicit, but I believe we dont need the friend class declarations.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
+    data_batch(uint64_t batch_id, std::unique_ptr<idata_representation> data);
+
+    uint64_t _batch_id;
+    std::unique_ptr<idata_representation> _data;
+  };
 
   /**
-   * @brief Destructor decrements processing count and potentially transitions state.
+   * @brief RAII read-only accessor. Borrows the parent, does not extend lifetime.
```
dhruv9vats: We _want_ to extend the lifetime.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
+   public:
+    const data_batch* operator->() const { return &(parent_->_batch); }
+    const data_batch& operator*() const { return parent_->_batch; }
 
-  // Move-only semantics
-  data_batch_processing_handle(const data_batch_processing_handle&)            = delete;
-  data_batch_processing_handle& operator=(const data_batch_processing_handle&) = delete;
+    /**
+     * @brief Downgrade from mutable → read-only.
+     * Internally releases unique lock, then blocks until shared lock acquired.
```
dhruv9vats: > then blocks until

I feel the user should be allowed to make this decision, and inspect if they can get a write lock quickly or not.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
-  success,
-  task_not_created,
-  invalid_state,
-  memory_space_mismatch,
-  missing_data,
-  not_attempted
-};
+    /**
+     * @brief Upgrade from read-only → mutable.
+     * Internally releases shared lock, then blocks until unique lock acquired.
```
dhruv9vats: Again, should be the users choice whether they want to wait or not.

dhruv9vats:
> idata_batch_probe, ... removed

We would need the probing ability for planned observability in the near future. We should think about how we want to do that.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
+  std::optional<mutable_data_batch> try_get_mutable();
 
-  /**
-   * @brief Blocking call: wait until the batch can accept a new task, then create it.
-   *
-   * Blocks until the batch leaves in_transit state, then performs the same transition
-   * as try_to_create_task(). Always succeeds when it returns.
-   */
-  void wait_to_create_task();
+  // -- Immutable field exposed on wrapper (lock-free, for repository lookups) --
```
dhruv9vats: The use of `const` for the `batch_id_` field in the `data_batch` should be considered.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,526 +41,198 @@ class memory_space;
...
-   * @brief Blocking call: wait until the batch can be locked for in-transit, then lock it.
-   *
-   * Blocks until processing_count == 0 and the batch is in idle or task_created state,
-   * then performs the same transition as try_to_lock_for_in_transit(). Always succeeds
-   * when it returns.
-   */
-  void wait_to_lock_for_in_transit();
+  bool subscribe();
+  void unsubscribe();
+  size_t get_subscriber_count() const;
```
dhruv9vats: Could you add some doc string to elaborate on the use case of these explicit counters? Is this something where the shared_ptr provided [use_count](https://en.cppreference.com/w/cpp/memory/shared_ptr/), and [unique](https://en.cppreference.com/w/cpp/memory/shared_ptr/unique.html) will not suffice? Or did you intend to keep counters of the number of read copies there are currently active or something like that?

dhruv9vats:
> One thing that I still don't know is the best API to pop `data_batch` from the repository with this new model. Previously we passed in the `batch_state` that we want to transition it into, which I think is not a good API but it allows us to skip `data_batch` that does not meet the state machine requirement to transition to that state.
> 
> For example, when popping data batch from the repo for downgrade it allows us to skip data batch that's currently in the `in_processing`.
> 
> But right now, since the state is gone, we will just pop `data_batch` from repository in a FIFO manner and it's up to the caller what do they want to do with it. The problem is now we might run into a 'pop and then insert again' scenarios where we pop a data batch, notice that we cannot downgrade cause a task is executing on it, and have to insert again to the repo.
> 
> Let me know if people have some ideas @aminaramoon @dhruv9vats

we can circumvent this by having an API on the data repository along the lines of `try_pop` that basically goes through all the batches it has in some order (maybe even a configurable order), and tries to obtain a write lock on each. The first batch on which a write lock is obtained is returned. So basically:

```cpp
std::optional<mutable_data_batch> try_pop() {
	std::optional<mutable_data_batch> result = std::nullopt;
	auto position = std::ranges::find_if(batches_, [&to_return](std::shared_ptr<synchronized_data_batch>& batch) { 			
		result = batch.try_get_mutable();
		return result != std::nullopt;
	});
	if (pos != batches_.end()) {
		std::erase(pos);
	}

	return result;
}
```
or something like this. And then have a blocking API that can be used if none of the batches can be readily downgraded.

mbrobbel:
/ok to test [acd69dd](https://github.com/NVIDIA/cuCascade/pull/99/commits/acd69dd80d8fe85d1b8d58c0665db84b3e8b73af)

mbrobbel:
/ok to test [b45fa20](https://github.com/NVIDIA/cuCascade/pull/99/commits/b45fa2060aaef27b6da3e73d9e35cc273928207e)

src/data/data_batch.cpp
```diff
@@ -16,455 +16,156 @@
...
-data_batch::data_batch(uint64_t batch_id,
-                       std::unique_ptr<idata_representation> data,
-                       std::unique_ptr<idata_batch_probe> probe)
-  : _batch_id(batch_id), _data(std::move(data)), _probe(std::move(probe))
+bool data_batch::subscribe()
 {
-  std::unique_lock<std::mutex> lock(_mutex);
-  update_state_to(batch_state::idle, lock);
+  _subscriber_count.fetch_add(1, std::memory_order_relaxed);
+  return true;
```
wmalpica, line 34: This always returns true, which is not what the function description in the header says. 
Do we even need this to return a bool? what is the intended purpose?

wmalpica (changes requested):

src/data/data_batch.cpp
```diff
@@ -16,455 +16,156 @@
...
-      ++_task_created_count;
-      should_notify = true;
-      success       = true;
-    }
-  }
-  if (should_notify) { _internal_cv.notify_all(); }
-  if (should_notify && cv_to_notify) { cv_to_notify->notify_all(); }
-  return success;
+  auto ptr = std::move(accessor._batch);
+  ptr->_state.store(batch_state::idle, std::memory_order_relaxed);
```
wmalpica: If we have multiple read_only_data_batch, then just converting one to idle should not set the sate to idle. 
Additionally, read_only_data_batch destructor should basically do the same thing as what this function is trying to do.

src/data/data_batch.cpp
```diff
@@ -16,455 +16,156 @@
...
-      cv_to_notify = _state_change_cv;
-    }
-    // Always increment and always notify: wait_to_lock_for_processing waits on
-    // _internal_cv checking _task_created_count > 0, so it must be woken here.
-    ++_task_created_count;
-  }
-  _internal_cv.notify_all();
-  if (cv_to_notify) { cv_to_notify->notify_all(); }
+  auto ptr = std::move(accessor._batch);
+  ptr->_state.store(batch_state::idle, std::memory_order_relaxed);
```
wmalpica: mutable_data_batch destructor should do what this is basically trying to do. The state needs to be reset upon the mutable_data_batch destruction

wmalpica (changes requested):

bwyogatama:
Replaced by #117

PR #103, opened 2026-04-13
Sync stream before destroying GPU representation in convert_to

mbrobbel:
## Summary
- Conversions from GPU in `data_batch::convert_to` may enqueue async operations on the provided CUDA stream that read from the source GPU memory. Previously, the source GPU representation was destroyed immediately after reassigning `_data`, risking use-after-free or leaking buffers into process-lifetime static storage.
- Synchronize the stream before the old GPU representation is destroyed to ensure async operations complete.
- Use `unique_lock` instead of `lock_guard` to release the mutex before synchronizing, avoiding blocking other threads during the stream sync.
- The sync condition triggers for any conversion **from** a GPU source, not just GPU→HOST — this covers future converters (e.g. GPU→DISK) and custom converters that may enqueue async reads without synchronizing internally.

## Test plan
- [x] New test: stream is synchronized before old GPU representation is destroyed
  - Passes with fix, fails without fix (`synced_before_destroy` = false)
- [x] New test: mutex is released before stream sync, allowing concurrent access
  - Passes with early unlock, fails when sync is inside the lock (`accessed_during_sync` = false)
  - Uses `cudaLaunchHostFunc` callback (50 ms sleep) for a deterministic window independent of GPU speed
- [x] Without any fix: first test fails (no sync), second test passes (no lock contention since no sync)
- [ ] Verify full test suite passes

### Local test results

| Configuration | Test 1 (sync before destroy) | Test 2 (mutex released during sync) |
|---|---|---|
| **With fix** (early unlock + sync) | ✅ Pass | ✅ Pass |
| **No fix** (original code) | ❌ Fail — no sync at all | ✅ Pass — no lock contention since no sync |
| **Sync inside lock** (naive fix) | ✅ Pass — sync happens | ❌ Fail — mutex held during 50 ms sync |

Only the correct implementation (early unlock + sync) passes both tests.

<details>
<summary>Raw test output</summary>

**With fix (early unlock + sync):**
```
Filters: [convert_to]
===============================================================================
All tests passed (3 assertions in 2 test cases)
```

**Without fix (original code, no sync at all):**
```
/home/matthijs/sirius/cucascade/test/data/test_data_batch.cpp:1753: FAILED:
  REQUIRE( observer.synced_before_destroy )
with expansion:
  false

test cases: 2 | 1 passed | 1 failed
assertions: 3 | 2 passed | 1 failed
```

**With sync inside lock (correct sync, but naive lock_guard):**
```
/home/matthijs/sirius/cucascade/test/data/test_data_batch.cpp:1849: FAILED:
  REQUIRE( accessed_during_sync )
with expansion:
  false

test cases: 2 | 1 passed | 1 failed
assertions: 2 | 1 passed | 1 failed
```

</details>

### How the tests work

**Test 1 — sync before destroy:** Registers a custom converter that enqueues a 4 MB async `cudaMemcpyDeviceToHost` into pinned memory and records a CUDA event, but does **not** synchronize. A custom `observed_gpu_representation` checks `cudaEventQuery` in its destructor. Without the sync, the event is still pending at destruction time.

**Test 2 — mutex released before sync:** Same converter setup, plus a `cudaLaunchHostFunc` callback that sleeps for 50 ms on the stream (creating a deterministic delay independent of GPU speed). A second thread calls `get_state()` (which acquires the mutex) and checks the event. With the early unlock, `get_state()` returns immediately while the 50 ms callback is still running. With the sync inside the lock, `get_state()` is blocked for the full 50 ms.

🤖 Created by Claude

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/include/cucascade/data/data_batch.hpp
+++ b/include/cucascade/data/data_batch.hpp
@@ -543,13 +543,15 @@ void data_batch::convert_to(representation_converter_registry& registry,
   _data = std::move(new_representation);
 
   bool needs_sync =
-    old_representation != nullptr && old_representation->get_current_tier() == memory::Tier::GPU;
+    old_representation != nullptr &&
+    (old_representation->get_current_tier() == memory::Tier::GPU ||
+     _data->get_current_tier() == memory::Tier::GPU);
 
   lock.unlock();
 
   if (needs_sync) {
-    // Conversions from GPU may enqueue async operations on the provided stream
-    // that read from the source GPU memory.  Synchronize before the old
+    // Conversions involving GPU may enqueue async operations on the provided
+    // stream that read from the source memory.  Synchronize before the old
     // representation is destroyed to avoid use-after-free.
     stream.synchronize();
   }
--- a/include/cucascade/data/data_batch.hpp
+++ b/include/cucascade/data/data_batch.hpp
@@ -542,16 +542,15 @@ void data_batch::convert_to(representation_converter_registry& registry,
   auto old_representation = std::move(_data);
   _data = std::move(new_representation);
 
-  bool needs_sync = old_representation != nullptr && target_memory_space != nullptr &&
-                    old_representation->get_current_tier() == memory::Tier::GPU &&
-                    target_memory_space->get_tier() == memory::Tier::HOST;
+  bool needs_sync =
+    old_representation != nullptr && old_representation->get_current_tier() == memory::Tier::GPU;
 
   lock.unlock();
 
   if (needs_sync) {
-    // GPU->HOST conversion may enqueue async copies on the provided stream.
-    // Make completion explicit before the source GPU representation is destroyed,
-    // instead of leaking it into process-lifetime static storage.
+    // Conversions from GPU may enqueue async operations on the provided stream
+    // that read from the source GPU memory.  Synchronize before the old
+    // representation is destroyed to avoid use-after-free.
     stream.synchronize();
   }
 }
--- a/include/cucascade/data/data_batch.hpp
+++ b/include/cucascade/data/data_batch.hpp
@@ -531,7 +531,7 @@ void data_batch::convert_to(representation_converter_registry& registry,
                             const cucascade::memory::memory_space* target_memory_space,
                             rmm::cuda_stream_view stream)
 {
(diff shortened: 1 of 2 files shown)


include/cucascade/data/data_batch.hpp
```diff
@@ -531,15 +531,29 @@ void data_batch::convert_to(representation_converter_registry& registry,
...
   }
 
   auto new_representation =
     registry.convert<TargetRepresentation>(*_data, target_memory_space, stream);
+  auto old_representation = std::move(_data);
   _data = std::move(new_representation);
+
+  bool needs_sync = old_representation != nullptr && target_memory_space != nullptr &&
+                    old_representation->get_current_tier() == memory::Tier::GPU &&
+                    target_memory_space->get_tier() == memory::Tier::HOST;
```
dhruv9vats: this might be too specific? that is, do we know that only conversions between GPU and HOST tiers will entail async copies? what about GPU to storage? also the other way around from other tiers to GPU? do we know these other combinations will not do any async copies, ever (like in case of a possible new representation added down the line)?

what if we stream synchronize if either the source or the destination tier is GPU?

dhruv9vats (commented):
Thanks for catching this!

include/cucascade/data/data_batch.hpp
mbrobbel: Makes sense! I'll update.

include/cucascade/data/data_batch.hpp
```diff
@@ -531,15 +531,28 @@ void data_batch::convert_to(representation_converter_registry& registry,
...
     throw std::runtime_error("Cannot convert representation while there is active processing");
   }
 
   auto new_representation =
     registry.convert<TargetRepresentation>(*_data, target_memory_space, stream);
+  auto old_representation = std::move(_data);
   _data = std::move(new_representation);
+
+  bool needs_sync =
+    old_representation != nullptr && old_representation->get_current_tier() == memory::Tier::GPU;
```
dhruv9vats: do we not want to cover the _to_ GPU memory case too?

felipeblazing (approved):

PR #105, opened 2026-04-15
Support static cudart linking

mbrobbel:
## Summary

Adds support for selecting the CUDA runtime library (shared vs static) at build time via the standard `CMAKE_CUDA_RUNTIME_LIBRARY` variable. This allows cuCascade to link against `cudart_static` when consumers require fully static CUDA runtime linkage (e.g. for self-contained deployments or to avoid `libcudart.so` version mismatches at runtime).

## Changes

- Derive `CUCASCADE_CUDART_TARGET` (`CUDA::cudart` or `CUDA::cudart_static`) from `CMAKE_CUDA_RUNTIME_LIBRARY`, defaulting to `Shared` when unset
- Replace the hardcoded `CUDA::cudart` in the public link interface with the selected target, so the choice propagates transitively to all consumers
- Set `CUDA_RUNTIME_LIBRARY` property on `cucascade_static` and `cucascade_shared` targets
- Remove redundant explicit `CUDA::cudart` from test and benchmark targets — they get it transitively via `cucascade`
- Error out with a clear message (listing accepted values) on unsupported `CMAKE_CUDA_RUNTIME_LIBRARY` values

## Usage

```cmake
# Static cudart (default is Shared)
cmake -DCMAKE_CUDA_RUNTIME_LIBRARY=Static ..
```

🤖 Created by Claude

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,8 @@ elseif(CUCASCADE_CUDA_RUNTIME_LIBRARY STREQUAL "Shared")
 else()
   message(
     FATAL_ERROR
-    "Unsupported CUDA runtime library: ${CUCASCADE_CUDA_RUNTIME_LIBRARY}"
+    "Unsupported CUDA runtime library: '${CUCASCADE_CUDA_RUNTIME_LIBRARY}'."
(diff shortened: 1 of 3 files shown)


mbrobbel:
I'll handle this differently.

PR #107, opened 2026-04-21
Filter non-functional network devices from topology discovery

pentschev:
Network devices discovered via `/sys/class/infiniband/` were being assigned to `UCX_NET_DEVICES` even when they were unusable (ports down, no IP configured, or missing `/dev/infiniband/uverbs*` device nodes in containers). This adds a `NetworkDeviceVerification` enum to `topology_discovery::discover()` that validates devices by checking port state, uverbs accessibility, and IP address presence, defaulting to the strictest level so non-functional devices are excluded automatically. Less strict levels are available for environments that use native RDMA verbs without IP-based transports (`EXISTS_ACTIVE`), or for diagnostic purposes where all visible hardware should be enumerated regardless of state (`EXISTS`) such as showing that the hardware exists/is installed.


mbrobbel (approved):

pentschev:
Thanks @mbrobbel and @felipeblazing !

PR #112, opened 2026-04-22
feat(data): add memory-space bandwidth profiler

felipeblazing:
Introduces measure_bandwidth() — a pure init-time function that measures pairwise transfer throughput between a set of memory_space objects via the converter registry. Produces an asymmetric bandwidth_profile that engines can consume for routing decisions.

* chunked_resource_info — opt-in mixin letting allocators that hand out fixed-size chunks advertise max_chunk_bytes() without altering the rmm::mr::device_memory_resource contract. fixed_size_host_memory_resource inherits it; memory_space exposes the probe via get_chunked_resource_info().
* bandwidth_profiler — enumerates GPU/HOST all-pairs plus disk<->{GPU,HOST} (no disk<->disk), measures both directions, records per-size samples and a median summary per pair. Uses converters so user-registered overrides are honored. Optional posix_fadvise(DONTNEED) between timed iterations evicts source files from the page cache so disk reads reflect real NVMe throughput instead of cached reads.
* pipeline_io_backend — replace the single shared copy_stream/order_event with a per-device cache, lazy-created on first use from each GPU context. Unblocks cross-context usage (e.g. GPU:N <-> DISK for N > 0) which previously failed with cudaErrorInvalidResourceHandle.
* Design doc at docs/bandwidth-profiler.md.

### DIFF
--- a/docs/bandwidth-profiler.md
+++ b/docs/bandwidth-profiler.md
@@ -1,0 +1,159 @@
+# Memory Space Bandwidth Profiler — Design
+
+**Status:** Design locked, pending implementation
+**Date:** 2026-04-21
+
+## Purpose
+
+Given a list of `memory_space*`, measure pairwise transfer throughput between them so engines can use the resulting bandwidth matrix for routing decisions (where to place / downgrade / upgrade data).
+
+Measurements flow through the **existing converter registry and `disk_io_backend`** — results reflect the actual transfer paths engines will pay at runtime, including user-registered converters.
+
+## Scope
+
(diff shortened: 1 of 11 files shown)


docs/bandwidth-profiler.md
```diff
@@ -0,0 +1,159 @@
+# Memory Space Bandwidth Profiler — Design
+
+**Status:** Design locked, pending implementation
```
wmalpica, line 3: NIT: pending implementation?

wmalpica (approved):

PR #117, opened 2026-04-28
implement 3-class data_back model and get rid of state machine

wmalpica:
This PR replaces this one https://github.com/NVIDIA/cuCascade/pull/99
Had to create a new clean branch to make the rebase easier





## Summary

Replace the 4-state finite state machine (`batch_state`: idle, task_created, processing, in_transit) with a 3-class design where `data_batch` is the "idle" state and all data access requires acquiring a lock through RAII accessor types.

**Core invariant:** it is impossible to read or mutate batch data without holding the appropriate lock, and move semantics make stale references a compile error.

## Design

```
  data_batch (idle — owns shared_mutex + data representation)                                                                                                                                    
  ├── Lock-free public API: get_batch_id(), subscribe(), unsubscribe(), get_subscriber_count(), get_state(), get_read_only_count()
  ├── Locking transitions (public, acquire mutex): to_read_only(), to_mutable(), try_to_read_only(), try_to_mutable()                                                                            
  ├── Private data accessors: get_data(), get_current_tier(), get_memory_space(), set_data()                                                                                                     
  │   └── Only accessible by friend accessor classes                                                                                                                                             
  │                                                                                                                                                                                              
  ├── read_only_data_batch (RAII shared lock, holds shared_ptr<data_batch>)                                                                                                                      
  │   └── Read accessors + clone() + clone_to()                                                                                                                                                  
  │                                                                                                                                                                                              
  └── mutable_data_batch (RAII exclusive lock, holds shared_ptr<data_batch>)                                                                                                                     
      └── Read accessors + clone() + clone_to() + set_data() + convert_to() 
```

### Observable state

`data_batch` exposes an atomic `batch_state` enum (`idle`, `read_only`, `mutable_locked`) via `get_state()`. Updated during every state transition. Enables the repository to filter batches by lock state.

### State transitions

**Non-static methods** (use `shared_from_this` — caller's `shared_ptr` is NOT consumed):
```cpp
auto accessor = batch->to_read_only();   // batch still valid, state = read_only
auto accessor = batch->to_mutable();     // batch still valid, state = mutable_locked
```

**Static methods** (consume accessors via move):
```cpp
auto batch = data_batch::to_idle(std::move(ro));         // ro is now consumed
```

**Locked-to-locked transitions** (static, consume via move):
```cpp
auto mut = data_batch::readonly_to_mutable(std::move(ro));   // upgrade lock
auto ro = data_batch::mutable_to_readonly(std::move(mut));   // downgrade lock
```

Non-blocking variants (`try_to_read_only`, `try_to_mutable`) return `std::optional`.

###  Simplified repository API

`pop_data_batch()` replaced with `pop_next_data_batch()`, a simple non-blocking FIFO pop that returns the first batch in the partition regardless of its lock state, returning  nullptr if the partition is empty:
```cpp                                                                                                                                                                                                 
auto batch = repo.pop_next_data_batch();   // first batch in partition, any state
```
                                                                                                                                                                                                 
  State filtering is no longer a repository concern. The caller acquires the appropriate lock via the accessor RAII types after retrieval. Because the pop no longer needs to wait for a batch to
   reach a target state, there is no blocking wait — the condition variable is no longer used internally and `pop_next_data_batch` always returns immediately.                                     
                                                                                                                                                                                                 
  `pop_data_batch_by_id()` and `get_data_batch_by_id()` were simplified in the same way: the `std::optional<batch_state> target_state` parameter was removed from both, and both return immediately    
  rather than blocking.
                                                                                                                                                                                                 
  Batches no longer hold a back-pointer to their containing repository's condition variable. `set_state_change_cv` was removed from `data_batch`, and the corresponding call was removed from        
  `add_data_batch.`

`pop_data_batch_by_id()` and `get_data_batch_by_id()` unchanged — caller checks state after retrieval.

## Breaking changes


  - data_batch now inherits std::enable_shared_from_this<data_batch>                                                                                                                             
  - data_batch has observable batch_state enum and get_state() method
  - batch_state enum replaced (old: idle/task_created/processing/in_transit → new: idle/read_only/mutable_locked)                                                                                
  - read_only_data_batch and mutable_data_batch are no longer class templates — they always hold std::shared_ptr<data_batch> (the unique_ptr variant was unused)                                 
  - data_batch now inherits std::enable_shared_from_this<data_batch>
  - data_batch has observable batch_state enum and get_state() method
  - batch_state enum replaced (old: idle/task_created/processing/in_transit → new: idle/read_only/mutable_locked)
  - read_only_data_batch and mutable_data_batch are no longer class templates — they always hold std::shared_ptr<data_batch> (the unique_ptr variant was unused)
  - Static entry-point methods removed (to_read_only(PtrType&&), to_mutable(PtrType&&), try_to_read_only(PtrType&), try_to_mutable(PtrType&)) — use non-static shared_from_this variants instead
  - convert_to moved from data_batch (private) to mutable_data_batch (public)
  - mutable_data_batch now has clone() and clone_to() (matching read_only_data_batch)
  - batch_state enum replaced (old: idle/task_created/processing/in_transit → new: idle/read_only/mutable_locked)
  - read_only_data_batch and mutable_data_batch are no longer class templates — they always hold std::shared_ptr<data_batch> (the unique_ptr variant was unused)
  - Static entry-point methods removed (to_read_only(PtrType&&), to_mutable(PtrType&&), try_to_read_only(PtrType&), try_to_mutable(PtrType&)) — use non-static shared_from_this variants instead
  - convert_to moved from data_batch (private) to mutable_data_batch (public)
  - mutable_data_batch now has clone() and clone_to() (matching read_only_data_batch)
  - Static entry-point methods removed (to_read_only(PtrType&&), to_mutable(PtrType&&), try_to_read_only(PtrType&), try_to_mutable(PtrType&)) — use non-static shared_from_this variants instead
  - convert_to moved from data_batch (private) to mutable_data_batch (public)
  - mutable_data_batch now has clone() and clone_to() (matching read_only_data_batch)
  - pop_data_batch(target_state) removed — use pop_next_data_batch() (non-blocking, state-agnostic FIFO pop)
  - pop_data_batch_by_id() and get_data_batch_by_id() no longer take target_state and no longer block
  - pop_data_batch(target_state) removed — use pop_next_data_batch() (non-blocking, state-agnostic FIFO pop)
  - pop_data_batch_by_id() and get_data_batch_by_id() no longer take target_state and no longer block
  - set_state_change_cv removed from data_batch — batches no longer hold a back-pointer to a repository condition variable
  - idata_batch_probe, data_batch_processing_handle, lock_for_processing_result/status removed
  - set_state_change_cv removed from data_batch — batches no longer hold a back-pointer to a repository condition variable
  - idata_batch_probe, data_batch_processing_handle, lock_for_processing_result/status removed
  - data_repository_manager::add_data_batch_impl uses if constexpr instead of SFINAE
  - data_repository_manager::add_data_batch_impl uses if constexpr instead of SFINAE

## Files changed

| Area | Files | What |
|------|-------|------|
| Core type system | `data_batch.hpp`, `data_batch.cpp` | 3-class rewrite, observable state, `enable_shared_from_this`, locked↔locked transitions, de-templatized accessors |
| Repository layer | `data_repository.hpp` | State-filtered `pop_idle/read_only/mutable_data_batch()` |
| Repository impl | `data_repository.cpp`, `data_repository_manager.cpp` | Updated type references |
| Converter | `representation_converter.cpp` | Added new batch type support |
| Tests | `test_data_batch.cpp`, `test_data_repository.cpp`, `test_data_repository_manager.cpp` | Full rewrite + new state/transition/filtered-pop tests |



wmalpica:
/ok to test https://github.com/NVIDIA/cuCascade/commit/4474bbcbe9474193daf401706337a763150ba31e

src/data/data_batch.cpp
```diff
@@ -16,455 +16,256 @@
...
-      success       = true;
-    }
+  if (_batch) {
+    // Decrement the reader count. If we were the last reader, transition to idle.
+    // NOTE: Do NOT call _lock.unlock() here — the _lock member destructor handles that.
+    // The destructor body runs before member destructors, so _batch is still valid here.
+    // After this function returns, _lock destructor fires first (declared after _batch,
+    // destroyed in reverse order), releasing the shared lock. Then _batch destructor fires.
+    auto prev = _batch->_read_only_count.fetch_sub(1);
+    if (prev == 1) { _batch->_state.store(batch_state::idle); }
```
bwyogatama, line 214: Just giving a note here that there is a brief moment where `batch_state = idle `but the lock is still held. The implication here is i guess during downgrade, when we are iterating for batch in idle state, it might return a batch that is still locked for a very brief amount of time. Not really a big deal but worth noting.

include/cucascade/data/data_batch.hpp
```diff
@@ -42,515 +39,467 @@ class memory_space;
...
-  data_batch_processing_handle() = default;
+  // -- Construction --
 
   /**
-   * @brief Construct a handle from a shared_ptr (stores weak_ptr; does not keep batch alive).
+   * @brief Construct a new data_batch.
    *
-   * @param batch shared_ptr to the data_batch to manage
+   * @param batch_id Unique identifier for this batch (immutable after construction).
+   * @param data     Owned data representation; must not be null.
```
bwyogatama, line 76: Should we enforce the `not be null` requirement as a contract?

src/data/data_batch.cpp
```diff
@@ -16,455 +16,256 @@
...
-  std::unique_lock<std::mutex> lock(_mutex);
-  return _task_created_count;
+  std::unique_lock<std::shared_mutex> lock(_rw_mutex, std::try_to_lock);
+  if (!lock.owns_lock()) { return std::nullopt; }
+  auto self = shared_from_this();
+  return mutable_data_batch(std::move(self), std::move(lock));
 }
 
-bool data_batch::try_to_cancel_task()
+// ========== Locked-to-locked static transitions ==========
```
bwyogatama, line 118: Can we put a comment somewhere that locked to locked transition is not atomic, and therefore another writer/reader technically can steal the lock during the transition

bwyogatama: I don't think there is a solution for this but a comment can be worth nothing, especially for AI when debugging issues

include/cucascade/data/data_batch.hpp
```diff
@@ -42,515 +39,467 @@ class memory_space;
...
+  // -- Named accessor methods (D-09/ACC-01) --
+
+  /** @brief Get the batch identifier. */
+  uint64_t get_batch_id() const { return _batch->get_batch_id(); }
+
+  /** @brief Get the memory tier of the held data. */
+  memory::Tier get_current_tier() const { return _batch->get_current_tier(); }
+
+  /** @brief Get a raw pointer to the data representation. */
+  idata_representation* get_data() const { return _batch->get_data(); }
```
bwyogatama, line 291: Is there a way to make this immutable? Cause right now technically someone can do `get_data` and mutate it with `read_only` lock

include/cucascade/data/data_batch.hpp
```diff
@@ -42,515 +39,467 @@ class memory_space;
...
+
+  /** @brief Get a raw pointer to the memory space. */
+  memory::memory_space* get_memory_space() const { return _batch->get_memory_space(); }
```
bwyogatama, line 294: Same comment, is there a way to make the `memory_space` immutable for `read_only`

include/cucascade/data/data_batch.hpp
bwyogatama: These raw pointers also espace the lock lifetime. Technically by getting this pointer, someone can modify/read the data even though the lock is already out of scope

bwyogatama (approved):

bwyogatama:
One more thing, I think the state machine documentation on data batch needs to be updated with this PR

felipeblazing (approved):

src/data/data_batch.cpp
wmalpica: Yeah, but the actual lock is still atomic. I dont this is worth worrying about

wmalpica:
/ok to test https://github.com/NVIDIA/cuCascade/commit/fcdeb3499c160bac7bba3dafccce2f39f798fcdc

include/cucascade/data/data_batch.hpp
wmalpica: I do like this idea, but perhaps we pursue this as a followup. I tried it out along with the get_memory_space and it involved A LOT of changes. I dont want to make this PR bigger still

PR #119, opened 2026-05-06
Use NVML to detect GPU NUMA node

nirandaperera:
GPU `numa_node` was reported as `-1` on hosts whose ACPI firmware (SRAT/SLIT tables) does not publish PCIe-to-NUMA affinity data. The kernel writes `-1` to `/sys/bus/pci/devices/<pci>/numa_node` in that case, and `topology_discovery::discover()` propagated that straight into `gpu_topology_info::numa_node` — even though the NVIDIA driver knows the correct NUMA node via its own PCI-bridge traversal (the same source `nvidia-smi topo -m` uses).

This change replaces the old mechanism with NVML-driven detection.

## Implementation notes

- `nodeSetSize = 1` (one `unsigned long` = 64 NUMA bits) — covers any realistic system; `CONFIG_NODES_SHIFT` defaults to 6 on x86_64 and the fallback only runs in the rare ACPI-quirk case anyway.
- `nvmlDeviceGetNumaNodeId` was considered but rejected: it returns the NUMA node *of the GPU itself*, which is only meaningful on Grace/GB200-style coherent platforms where the GPU is its own NUMA node.


src/memory/topology_discovery.cpp
```diff
@@ -46,7 +47,14 @@ struct NvmlLoader {
   nvmlReturn_t (*p_nvmlDeviceIsMigDeviceHandle)(nvmlDevice_t, unsigned int*)    = nullptr;
   nvmlReturn_t (*p_nvmlDeviceGetDeviceHandleFromMigDeviceHandle)(nvmlDevice_t,
                                                                  nvmlDevice_t*) = nullptr;
-  const char* (*p_nvmlErrorString)(nvmlReturn_t)                                = nullptr;
+  // Used as a fallback when ACPI SRAT lacks PCIe affinity data and sysfs reports
+  // numa_node = -1. Required: NVML has shipped this since the CUDA 8.0 driver
+  // branch (~2016) and it is documented as available on all Kepler+ devices.
```
pentschev: Nobody 
```suggestion
  // numa_node = -1.
```

pentschev: Nobody cares to know about Kepler or CUDA 8.0 anymore.

pentschev: Actually remove the entire comment, the first part is duplicated in line 873, where it is more appropriate.

src/memory/topology_discovery.cpp
```diff
@@ -857,7 +869,19 @@ bool topology_discovery::discover(NetworkDeviceVerification net_verification)
...
+      // topology rather than relying on firmware tables, so it usually has the
+      // correct answer (same source as `nvidia-smi topo -m`).
+      if (gpu.numa_node == -1) {
+        unsigned long nodeset = 0;
+        if (nvml.p_nvmlDeviceGetMemoryAffinity(device, 1, &nodeset, NVML_AFFINITY_SCOPE_NODE) ==
+              NVML_SUCCESS &&
+            nodeset != 0) {
+          gpu.numa_node = std::countr_zero(nodeset);
+        }
+      }
```
pentschev: I'm actually considering whether we need `get_numa_node_from_sys` anyway, do we use it elsewhere? If not, we should probably just remove it completely and rely directly on `nvmlDeviceGetMemoryAffinity` for this.

test/memory/test_topology_discovery.cpp
```diff
@@ -182,6 +182,37 @@ TEST_CASE("Topology Discovery default uses strictest verification", "[hw_topolog
...
+// regression that re-introduced the -1 leak.
+TEST_CASE("Topology Discovery resolves GPU NUMA node on NUMA-aware hosts", "[hw_topology]")
+{
+  topology_discovery discovery;
+  REQUIRE(discovery.discover());
+
+  auto const& topology = discovery.get_topology();
+
+  if (topology.num_gpus == 0 || topology.num_numa_nodes <= 0) {
+    SUCCEED("Skipped: requires at least one GPU and a NUMA-aware host");
```
pentschev: Does a non-NUMA-aware host even exists, at least in the context that we care about and test? I think not and we should remove the second condition.

test/memory/test_topology_discovery.cpp
```diff
@@ -182,6 +182,37 @@ TEST_CASE("Topology Discovery default uses strictest verification", "[hw_topolog
...
+    return;
+  }
+
+  for (auto const& gpu : topology.gpus) {
+    INFO("GPU " << gpu.id << " (" << gpu.name << " @ " << gpu.pci_bus_id << ")");
```
pentschev, line 200:
```suggestion
```

pentschev (changes requested):

src/memory/topology_discovery.cpp
nirandaperera: Sure, I was thinking the same TBH.

test/memory/test_topology_discovery.cpp
nirandaperera: Let's leave this. In catch2 this is like a scoped trace. So, no harm keeping this. It wont be printed unless the test fails

test/memory/test_topology_discovery.cpp
```diff
@@ -182,6 +182,37 @@ TEST_CASE("Topology Discovery default uses strictest verification", "[hw_topolog
...
 
+// Regression test for the bug where GPU NUMA node was reported as -1 on hosts whose ACPI
+// SRAT/SLIT tables do not publish PCIe-to-NUMA affinity data. The previous implementation
+// read /sys/bus/pci/devices/<pci>/numa_node, which returns -1 in that case;
+// topology_discovery now resolves the NUMA node via nvmlDeviceGetMemoryAffinity, which
+// walks the GPU driver's PCI bridge topology and is unaffected by firmware quirks.
+//
+// Invariant: when the host advertises NUMA topology and GPUs are present, every discovered
+// GPU must resolve to a valid NUMA node. This test would catch a regression that
+// re-introduced the -1 leak.
```
pentschev:
```suggestion
// Invariant: when the host advertises NUMA topology and GPUs are present, every discovered
// GPU must resolve to a valid NUMA node.
```

pentschev: We don't care about details of the previous implementation. Only the expected behavior, which is what being tested, matters.

src/memory/topology_discovery.cpp
```diff
@@ -323,25 +332,31 @@ std::vector<size_t> resolve_visible_gpu_indices(
...
- * @brief Get NUMA node from /sys for a PCI device.
+ * @brief Get the host NUMA node id with the best memory affinity to a GPU.
  *
- * Reads /sys/bus/pci/devices/<pci>/numa_node. If the file is missing, empty, or
- * cannot be parsed as an integer, returns -1.
+ * Queries NVML's `nvmlDeviceGetMemoryAffinity` with `NVML_AFFINITY_SCOPE_NODE` and
+ * returns the lowest-numbered NUMA node in the resulting bitmask. NVML walks the
+ * GPU driver's PCI bridge topology directly, so this is unaffected by ACPI
+ * SRAT/SLIT firmware quirks that cause `/sys/bus/pci/devices/<pci>/numa_node` to
+ * report -1 on otherwise NUMA-aware hosts. Same source as `nvidia-smi topo -m`.
```
pentschev:
```suggestion
 * Queries NVML's `nvmlDeviceGetMemoryAffinity` with `NVML_AFFINITY_SCOPE_NODE` and
 * returns the lowest-numbered NUMA node in the resulting bitmask. Same source as
 * `nvidia-smi topo -m`.
```

pentschev: Same here, we don't care about the previous implementations.

pentschev (changes requested):
Couple of request to tidy up docs/comments, after that we should be good.

test/memory/test_topology_discovery.cpp
pentschev: This hasn't been addressed in full, you left the "This test would catch a regression that re-introduced the -1 leak." sentence that refers to the old implementation. Please remove that.

pentschev (changes requested):
Please address https://github.com/NVIDIA/cuCascade/pull/119#discussion_r3319750045 and fix linting.

test/memory/test_topology_discovery.cpp
nirandaperera: I'm sorry missed that one.

pentschev (approved):
Thanks @nirandaperera . I'll get this merged now as I'm investigating something where this could be relevant.

PR #120, opened 2026-05-08
Tighten `data_batch`

dhruv9vats:
Proposes tightening the API on and around the `data_batch`.

These proposals are motivated by efforts to:
- Ensure correctness
- Highlight conflicting requirements
- Reduce duplication of code
- Separate concerns where possible
- Reduce escape hatches

listed approximately in the order of importance given while designing this.

This is an explicit "show me the code" _proposal_ implemented to gather concrete criticism to allow settling on a quality API for the central `data_batch` construct. There are comments for the reviewer that details why a specific change was made.


include/cucascade/data/data_batch.hpp
```diff
@@ -47,46 +47,37 @@ namespace cucascade {
...
- public:
-  // -- Construction --
-
-  /**
-   * @brief Construct a new data_batch.
-   *
-   * @param batch_id Unique identifier for this batch (immutable after construction).
-   * @param data     Owned data representation; must not be null.
-   */
-  data_batch(uint64_t batch_id, std::unique_ptr<idata_representation> data);
```
dhruv9vats, line 80: It is incorrect to 
```
data_batch : public std::enable_shared_from_this<data_batch>
``` 
_and_ have a public constructor. To be able to use `shared_from_this()`, the object must already be owned by a `shared_ptr`, otherwise it throws:
See https://en.cppreference.com/cpp/memory/enable_shared_from_this (specially example `misuseGood()`).

We rely heavily and primarily on `shared_from_this()` to extend lifetimes. Thus we must use a factory fn to create data batches and make the constructor private.

include/cucascade/data/data_batch.hpp
```diff
@@ -97,169 +88,225 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
+      // Conversions involving GPU may enqueue async operations on the provided
+      // stream that read from the source memory.  Synchronize before the old
+      // representation is destroyed to avoid use-after-free.
+      stream.synchronize();
+    }
+  }
 
-  // -- Static transition methods (D-13/D-14/D-15/D-17) --
+ private:
+  data_batch_core(uint64_t batch_id, std::unique_ptr<idata_representation> data);
```
dhruv9vats, line 234: This is private so only `data_batch` can create this core, and thus access is only through the RAII handles.

include/cucascade/data/data_batch.hpp
```diff
@@ -97,169 +88,225 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
+};
+
+/**
+ * @brief Synchronized data batch type allowing thread-safe access to the core data batch
+ * via RAII accessors.
+ */
+class data_batch : public std::enable_shared_from_this<data_batch> {
+  friend read_only_data_batch;
+  friend mutable_data_batch;
+
```
dhruv9vats, line 247: The primary motivations to separate `data_batch` and `data_batch_core` are:
1. Reduce code duplication, so no more:
```cpp

Tier read_only_data_batch::get_current_tier() { return _batch->get_current_tier(); }
Tier  mutable_data_batch::get_current_tier() { return _batch->get_current_tier(); }

memory_space* read_only_data_batch::get_memory_space() { return _batch->memory_space(); }
memory_space*  mutable_data_batch::get_memory_space() { return _batch->memory_space(); }
```
And the convert_to / clone_to template declarations.

The RAII handles instead simply act as "stencils", conditionally providing the access/"view" to `data_batch_core` APIs via operator overloading, instead of a middlemen role.
 
2. Allows separating concerns
- This I don't have a strong opinion on. But this structure allows keeping the `data_batch_core` basically as a POD, and move the thread-safety concerns to the parent `data_batch`

3. This is the class where future `data_batch_probe` like constructs may be added.

include/cucascade/data/data_batch.hpp
```diff
@@ -269,78 +316,48 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
   read_only_data_batch(read_only_data_batch&& other) noexcept;
   read_only_data_batch& operator=(read_only_data_batch&& other) noexcept;
-  ~read_only_data_batch();
+  read_only_data_batch(const read_only_data_batch& other)            = delete;
+  read_only_data_batch& operator=(const read_only_data_batch& other) = delete;
+
+  // will allow only read access
+  const data_batch_core* operator->() const { return &(_owner->_batch); }
+  const data_batch_core& operator*() const { return _owner->_batch; }
+
```
dhruv9vats: Simply forwards to `data_batch_core`

include/cucascade/data/data_batch.hpp
```diff
@@ -269,78 +316,48 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
+    }  // destroy accessor, releasing shared lock
+    return ptr;
+  }
 
-  // -- Copy support: acquires a new shared lock, increments reader count --
-  read_only_data_batch(const read_only_data_batch& other);
-  read_only_data_batch& operator=(const read_only_data_batch& other);
+  read_only_data_batch clone_read_only_access() const
+  {
+    std::optional<read_only_data_batch> maybe_clone = _owner->try_get_read_only();
```
dhruv9vats: Explicitness here felt better instead of a copy constructor. So the caller knows exactly what is being copied. though no strong opinion here either.

include/cucascade/data/data_batch.hpp
```diff
@@ -367,162 +384,43 @@ class read_only_data_batch {
...
-  /** @brief Get the memory tier of the held data. */
-  memory::Tier get_current_tier() const { return _batch->get_current_tier(); }
-
-  /** @brief Get a raw pointer to the data representation. */
-  idata_representation* get_data() const { return _batch->get_data(); }
-
-  /** @brief Get a raw pointer to the memory space. */
-  memory::memory_space* get_memory_space() const { return _batch->get_memory_space(); }
-
-  // -- Write methods (D-10/ACC-02) --
```
dhruv9vats: > (D-10/ACC-02)

not sure what these are, indicators of some sort? LLM bread crumbs?

include/cucascade/data/data_repository_manager.hpp
```diff
@@ -127,8 +127,7 @@ class data_repository_manager {
   /**
    * @brief Add a data_batch to specified operator repositories.
    *
-   * For shared_ptr: The batch is copied to each repository.
-   * For unique_ptr: This method requires only one operator (single owner semantics).
```
dhruv9vats, line 131: Under the assumption that we want `data_batch` to be thread-safe, these I believe end up being conflicting requirements, or atleast can't be supported by the same `data_batch`.

I dont think we can have both
- lifetime extending RAII handles that allow parallel reads and exclusive writes,
- _and_ have a `data_batch` under `unique_ptr` because`shared_from_this` then throws.

To support this we would at least need a wrapper over the `data_batch`, or another wrapper around `data_batch_core`.

dhruv9vats:
/ok to test [bffa10d](https://github.com/NVIDIA/cuCascade/pull/120/commits/bffa10d9c6dbed49423811069fe389927b878084)

include/cucascade/data/data_batch.hpp
```diff
@@ -97,169 +88,225 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
-   * Releases the shared lock, then acquires an exclusive lock (may block).
-   * The source accessor is consumed via move.
-   * NOTE: The transition is not atomic. 
+   * Atomic, lock-free.
    *
-   * @param accessor Rvalue reference to the read-only accessor (consumed).
-   * @return A mutable_data_batch holding the exclusive lock.
+   * @throws std::runtime_error if subscriber count is already zero.
    */
-  [[nodiscard]] static mutable_data_batch readonly_to_mutable(read_only_data_batch&& accessor);
```
dhruv9vats, line 215: since this might block, just having a single path `read_only -> idle -> get_mutable / try_get_mutable` that retains the choice of whether to wait or just try might be simpler?

include/cucascade/data/data_batch.hpp
```diff
@@ -97,169 +88,225 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
-   * @brief Decrement the subscriber interest count.
-   *
-   * Atomic, lock-free.
-   *
-   * @throws std::runtime_error if subscriber count is already zero.
+   * @brief Get a raw pointer to the data representation.
+   * @return Non-owning pointer to the data, or nullptr if empty.
    */
-  void unsubscribe();
+  idata_representation* get_data() const;
```
dhruv9vats: Maybe this and get_memory_space are made to return `const idata_representation*` ? if mutability of the underlying pointer content is not needed anywhere?

include/cucascade/data/data_repository_manager.hpp
```diff
@@ -246,22 +245,8 @@ class data_repository_manager {
...
-          "Use shared_ptr for multiple destinations.");
-      }
-      if (!ops.empty()) {
-        auto& op = ops[0];
-        _repositories[{op.first, std::string(op.second)}]->add_data_batch(std::move(batch));
-      } else {
-        throw std::runtime_error("No operator ports provided");
-      }
+    for (auto& op : ops) {
+      _repositories[{op.first, std::string(op.second)}]->add_data_batch(batch);
```
dhruv9vats, line 249: If the primary use case of the `unique_ptr<data_batch>` was to enforce the "can be added on only one repo", that might be instead handled at the repo/repo-manager level. If it was a consequence, we should think of use cases where we need to guarantee unique _ownership_, and the shared ownership model does not suffice. (Sirius current, AFAIK, does not make use of unique_(data_batch / repo / repo_manager))

docs/data-management.md
```diff
@@ -488,17 +475,18 @@ Application
...
     |       |-- returns read_only_data_batch (blocks until available)
-    |       |-- ro.get_data() grants read access to idata_representation
+    |       |-- ro->get_data() grants read access to idata_representation
     |
     |-- [on memory pressure]
-    |   mutable_data_batch mut = data_batch::readonly_to_mutable(std::move(ro))
-    |   mut.convert_to<TargetRepresentation>(registry, target_space, stream)
+    |   read_only_data_batch::to_idle(std::move(ro))      [release shared lock]
+    |   mutable_data_batch mut = batch->get_mutable()     [exclusive lock]
+    |   mut->convert_to<TargetRepresentation>(registry, target_space, stream)
```
wmalpica, line 485: Why do you have `mut->convert_to` instead of `mut.convert_to`? its not a pointer or an optional. 

Similarly you have several places where you also have `ro->`

include/cucascade/data/data_batch.hpp
```diff
@@ -97,169 +88,225 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
...
+   * @tparam TargetRepresentation Target representation type.
+   * @param registry           Converter registry for type-keyed dispatch.
+   * @param new_batch_id       Batch ID for the cloned batch.
+   * @param target_memory_space Target memory space for the converted data.
+   * @param stream              CUDA stream for memory operations.
+   * @return A new data_batch wrapped in shared_ptr.
    */
-  memory::memory_space* get_memory_space() const;
+  template <typename TargetRepresentation>
+  std::shared_ptr<data_batch> clone_to(representation_converter_registry& registry,
```
wmalpica: the data_batch class should not be able to make clones. That can only be done by the mutable_data_batch and read_only_data_batch classes

include/cucascade/data/data_batch.hpp
wmalpica: i am not in disagreement, but that will add a whole bunch more changes in Sirius. Not sure if its worth it

include/cucascade/data/data_repository_manager.hpp
wmalpica: The refactor that introduced read_only_data_batch and mutable_data_batch basically made it so that we cant have unique data_batch, and I think we decided that was ok.  Any code or comments that are still here that indicate that we could have unique data_batch was a mistake

wmalpica (commented):
In general i like the data_batch_core concept, and how much comment and documentation cleanup this PR brings. 
But requiring a make function, using get_read_only vs to_read_only and no copy constructors for read_only_data_batch, plus other changes means that this PR would require a very large number of changes in Sirius. I would recommend postponing that Sirius refactor until after some of the large initiatives currently underway are complete

dhruv9vats:
/ok to test [1566c28](https://github.com/NVIDIA/cuCascade/pull/120/commits/1566c283d36c9dda1312e357834da819e66d9df8)

include/cucascade/data/data_repository_manager.hpp
dhruv9vats: This is very useful info, thanks.

include/cucascade/data/data_batch.hpp
dhruv9vats: Agreed, definitely something that can be deferred or reverted.

docs/data-management.md
dhruv9vats: Great observation! The reason for that is, these RAII handles now delegate to the data_batch_core via the overloading of operators, like for read only:
```cpp
// will allow only read access
const data_batch_core* operator->() const { return &(_owner->_batch); }
const data_batch_core& operator*() const { return _owner->_batch; }
```

and not via forwarding member functions. So there's only a single definition of a function, and the end user just uses that, just _via_ that handles. And these RAII handles don't have any public member variables and will have minimal member functions.

include/cucascade/data/data_batch.hpp
dhruv9vats: that makes sense, addressed in [move clone and clone_to to data_batch_core](https://github.com/NVIDIA/cuCascade/pull/120/commits/1566c283d36c9dda1312e357834da819e66d9df8)

dhruv9vats:
/ok to test [b6370b6](https://github.com/NVIDIA/cuCascade/pull/120/commits/b6370b61b4957e13401f9dbdf53ad461e4d36bdc)

docs/data-management.md
wmalpica: I see.... I find that confusing and a leaky abstraction. I feel like data_batch_core should be an implementation detail designed to make it easier to maintain data_batch and read_only_data_batch and mutable_data_batch, but I think it should be a private detail. Using it this way exposes this internal implementation detail

dhruv9vats: >data_batch_core should be an implementation detail designed to make it easier to maintain data_batch and read_only_data_batch and mutable_data_batch

That is the idea with this, so there's only a single source of truth for what can be done with/on a data_batch -- the public members exposed via the data_batch_core. Its just that the only way to get to them would via the RAII handle. Because in this the the data_batch_core becomes a POD, and only holds onto the core underlying and is not concerned about thread safety or concurrent access, which, IMHO, makes is easier to understand what the data_batch_core can actually do.

> but I think it should be a private detail

If we make it private, we are again at wrapping/duplicating the underlying data_batch_core methods. The RAII handles basically act as stencils, providing guided access without being concerned with about what all methods are const vs mutable, simply passing it on. I am thinking of this as constrained access to the underlying object, as opposed to a leaky-abstraction, because, the underlying data is what we want to ultimately access/modify.

dhruv9vats:
/ok to test [e6db4f8](https://github.com/NVIDIA/cuCascade/pull/120/commits/e6db4f8db55195e31b0b82cac2cf1b4000bea057)

wmalpica (approved):

dhruv9vats:
/ok to test [c9cd306](https://github.com/NVIDIA/cuCascade/pull/120/commits/c9cd30675db28866474d480b971ed3177bab17b0)

dhruv9vats:
/ok to test [f6529c5](https://github.com/NVIDIA/cuCascade/pull/120/commits/f6529c591e49e1c76cf5e4edd617b2dd0085751f)

dhruv9vats:
/ok to test [f6529c5](https://github.com/NVIDIA/cuCascade/pull/120/commits/f6529c591e49e1c76cf5e4edd617b2dd0085751f)

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
+   * "f lock is called by a thread that already owns the shared_mutex in any mode (exclusive or
+   * shared), the behavior is undefined."
+   *
+   * Re-acquiring AFTER the prior accessor has been released is fine.
+   *
+   * A held accessor must not be moved to another thread and released there;
+   * from: https://en.cppreference.com/cpp/thread/shared_mutex/unlock:
+   * "The mutex must be locked by the current thread of execution, otherwise, the behavior is
+   * undefined. "
+   *
```
dhruv9vats, line 307: Some shared_mutex undefined behaviors we were subtly relying upon.

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
+   * @return The current memory tier.
+   */
+  [[nodiscard]] memory::Tier get_current_tier() const;
 
-  // -- Lock-free public API --
+  /**
+   * @brief Get a raw pointer to the data representation.
+   * @return Non-owning pointer to the data, or nullptr if empty.
+   */
+  [[nodiscard]] idata_representation* get_data() const;
```
mbrobbel, line 107: ﻿`get_data()` is `const` but returns a non-const pointer `idata_representation*`. A `read_only_data_batch` (shared lock) can do `ro->get_data()->release_table()/rebind_stream()/record_writer_event()` or a non-const `cast<T>()`, mutating the representation under a read lock. Two concurrent readers calling `record_writer_event()` race on the non-atomic `_writer_event` in `gpu_data_representation.cpp:117`. Maybe return `const idata_representation*` (+ a non-const overload reachable only via `mutable_data_batch`); And same for `get_memory_space()`.

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
+  data_batch_core* operator->() const
+  {
+    assert_validity();
+    return &(_owner->_batch);
+  }
+  data_batch_core& operator*() const
+  {
+    assert_validity();
+    return _owner->_batch;
+  }
```
mbrobbel, line 541: `set_data`/`get_data`/etc. are public, and `operator*/`-> return `&(_owner->_batch)`, which outlives the accessor.

```
auto& core = *batch->get_mutable();
core.set_data(...)
``` 
compiles, but releases the lock at the `;`, then mutates unlocked.

This means the comments (line 64 and 95) are inaccurate, to fix maybe forward guarded methods through the accessor so access is lifetime-coupled to the lock﻿.

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
+   *
+   * @note The returned owner pointer must NOT be used to acquire another accessor on the thread
+   * that already holds this one: accessors are NON-RECURSIVE and recursive locking of
+   * std::shared_mutex is undefined behavior. Re-acquiring AFTER this accessor is released is fine.
+   * A held accessor must also not be moved to another thread and released there (cross-thread
+   * unlock_shared is undefined behavior).
+   * Throws std::logic_error if the accessor handle was moved-from or reset.
    */
-  [[nodiscard]] cudaEvent_t get_writer_event() const
+  [[nodiscard]] std::shared_ptr<data_batch> get_owner_copy() const
```
mbrobbel, line 457: The note warns that re-acquiring an accessor on a thread that holds one is `shared_mutex` UB — but this method hands the caller a lockable owner `shared_ptr` while the lock is still held, which is exactly what's needed to do `o->get_read_only()` (recursive UB) or `o->get_mutable()` (self-deadlock). The warning can't be honored when the API itself supplies the footgun. Suggest yielding the owner only through the consuming `reset_and_release_owner` (which drops the lock first).

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
-  uint64_t get_batch_id() const;
+  [[nodiscard]] memory::memory_space* get_memory_space() const;
 
   /**
-   * @brief Increment the subscriber interest count.
+   * @brief Replace the data representation.
+   * @param data New data representation (takes ownership).
    */
-  void subscribe();
+  void set_data(std::unique_ptr<idata_representation> data) { _data = std::move(data); }
```
mbrobbel, line 119: This has no `null` check, so `_data` can become null — line 111 ("nullptr if data is null") and line 184 ("@throws std::runtime_error if the data is null") are not honored: `get_memory_space()` and `clone()` dereference unconditionally (both are real regressions vs main, which guarded them).

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
-  // -- Move-only --
-  mutable_data_batch(mutable_data_batch&& other) noexcept;
-  mutable_data_batch& operator=(mutable_data_batch&& other) noexcept;
-  ~mutable_data_batch();
-  mutable_data_batch(const mutable_data_batch&)            = delete;
-  mutable_data_batch& operator=(const mutable_data_batch&) = delete;
-
- private:
-  friend class data_batch;
+  void assert_validity() const
```
mbrobbel, line 587: `read_only_data_batch` documents and asserts "a valid lock is held if _owner exists"; `mutable_data_batch::assert_validity()` checks only `_owner` and omits `assert(_lock.owns_lock())`. Add it for symmetry.

src/data/data_batch.cpp
```diff
@@ -16,263 +16,188 @@
...
+    const size_t prev = _owner->_read_only_count.fetch_sub(1);
+    if (prev == 1) { _owner->_state.store(batch_state::idle); }
 
-std::shared_ptr<data_batch> read_only_data_batch::clone(uint64_t new_batch_id,
-                                                        rmm::cuda_stream_view stream) const
-{
-  if (_batch->_data == nullptr) { throw std::runtime_error("Cannot clone: data is null"); }
-  auto cloned_data = _batch->_data->clone(stream);
-  return std::make_shared<data_batch>(new_batch_id, std::move(cloned_data));
+    _lock.unlock();
```
mbrobbel, line 155: `_state`/`_read_only_count` updated non-atomically under a shared lock.

 No data corruption now (the mutex still guards real access, and nothing in `src/` reads `get_state()` for decisions), but it's a footgun for future consumers. Cleanest fix: derive `batch_state` from `_read_only_count` + an exclusive-held flag and drop the separate `_state` atomic.

src/data/data_batch.cpp
```diff
@@ -16,263 +16,188 @@
...
-                                                      rmm::cuda_stream_view stream) const
-{
-  if (_batch->_data == nullptr) { throw std::runtime_error("Cannot clone: data is null"); }
-  auto cloned_data = _batch->_data->clone(stream);
-  return std::make_shared<data_batch>(new_batch_id, std::move(cloned_data));
+    // INVARIANT: if the owner exists, a valid lock is held.
+    assert(_lock.owns_lock());
+
+    _owner->_state.store(batch_state::idle);
+    _lock.unlock();
```
mbrobbel, line 198: `mutable_data_batch::reset()` stores idle before `unlock()`

There's a window where `get_state()==idle` while the exclusive lock is still held. Folds into the same fix (derive state from `lock`/`count`, or store after unlock).

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
+  {
+    auto new_representation =
+      registry.convert<TargetRepresentation>(*_data, target_memory_space, stream);
+    auto old_representation = std::move(_data);
+    _data                   = std::move(new_representation);
+
+    bool needs_sync = (old_representation != nullptr &&
+                       old_representation->get_current_tier() == memory::Tier::GPU) ||
+                      _data->get_current_tier() == memory::Tier::GPU;
+
```
mbrobbel, line 146: When the old tier ≠ GPU, the `||` short-circuit evaluates `_data->get_current_tier()` on the freshly-assigned pointer; a converter returning `nullptr` makes this a null deref. Guard with `_data != nullptr &&`, and/or throw before assigning if convert returned `null`.

include/cucascade/data/data_batch.hpp
```diff
@@ -49,521 +52,571 @@ namespace cucascade {
...
 
   /**
-   * @brief Try to transition from idle to read-only (non-blocking).
+   * @brief Try to get a read read-only access handle (non-blocking).
+   *
+   * @note NON-RECURSIVE: do not call this on a thread that already holds any accessor for the
+   * same batch -- recursive locking of std::shared_mutex is undefined behavior. Re-acquiring
+   * AFTER the prior accessor has been released is fine. A held accessor must not be moved to
+   * another thread and released there. (Debug builds detect same-thread recursion via an
+   * ownership guard.)
```
mbrobbel, line 319: No thread-ownership tracking exists — `assert(_lock.owns_lock())` can't detect recursion (a fresh `try_to_lock` lock reports `owns_lock()==true`, and it isn't even checked at acquisition). This actively encourages the misuse it claims to catch. Either remove the sentence or implement a real debug-only per-thread holder check

mbrobbel (commented):
Some feedback collected with claude and codex. Also note that `docs/data_batch_state_transitions.md` / `docs/data-management.md` / `docs/ARCHITECTURE.md` should be updated with the new API.

And `test/data/test_data_batch.cpp:921` and `:1016` — `convert_to sync` tests both detect a missing `stream.synchronize()` only via a 4 MB async copy still being in flight, which finishes too fast on a warm GPU — so they'd pass even if the sync were removed. The third test (`:1028`) is
deterministic because it enqueues a 50 ms `cudaLaunchHostFunc` delay; apply the same pattern to these two so the negative case is actually caught.

And `test/data/test_data_batch.cpp:555` — DISK-tier clone test asserts a false invariant. Cloning succeeds only because `mock_data_representation::clone()` copies itself; the real `disk_data_representation::clone()` always throws (documented contract). The test implies `data_batch` clone works for disk
batches when it doesn't — either drop the DISK section or assert it throws.

dhruv9vats:
Superseded by #154 and others.

PR #123, opened 2026-05-12
Stop enabling C for C++ and CUDA builds

pentschev:
cuCascade only builds C++ and CUDA sources, so declaring C as a project language forces downstream builds to initialize and validate an unused compiler. That can break embedded package builds when the parent project has already initialized its C++ and CUDA toolchain state and a later C compiler check inherits incompatible launcher settings.

Drop the unused C language requirement so consumers only need the toolchains cuCascade actually uses. This keeps standalone builds working while avoiding unnecessary compiler checks in downstream integrations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ cmake_minimum_required(VERSION 3.26.4 FATAL_ERROR)
 project(
   cuCascade
   VERSION 0.1.0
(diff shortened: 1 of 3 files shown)


felipeblazing (approved):

pentschev:
Thanks Felipe!

PR #126, opened 2026-05-13
Split topology discovery into its own target and switch to static CUDA linking

KyleFromNVIDIA:
rapidsmpf's rrun utility uses topology discovery without any of cucascade's other functionality, and thus does not require its dependencies. Expose a separate library target that rrun can use.

Additionally switch to static CUDA linking to match RAPIDS and broader CUDA recommendations.


pentschev:
I'm +1 on this. For additional context to @felipeblazing et al.: without a separate target, linking to `libcucascade.so` directly forces us to also link to `libcudf`, then transitively `libkvikio`, `libnvcomp`, etc. This becomes a problem because we need to have executables that use the topology discovery also link to those libraries, and for wheels specifically this is more problematic since those DSOs are visible to Python but not C/C++ binaries.

test/CMakeLists.txt
pentschev, line 1: Ok, `test/data`/`test/memory`/`test_utils` implementations are all explicitly added in this file and the `CMakeLists.txt` are unused, so should be safe to remove them.

CMakeLists.txt
```diff
@@ -136,20 +143,42 @@ target_link_libraries(cucascade_objects PUBLIC ${CUCASCADE_PUBLIC_LINK_LIBS})
...
+
   add_library(cucascade_static STATIC $<TARGET_OBJECTS:cucascade_objects>)
   add_library(cuCascade::cucascade_static ALIAS cucascade_static)
 
   target_include_directories(cucascade_static
                              PUBLIC ${CUCASCADE_PUBLIC_INCLUDE_DIRS})
-  target_link_libraries(cucascade_static PUBLIC ${CUCASCADE_PUBLIC_LINK_LIBS})
+  target_link_libraries(
+    cucascade_static PUBLIC ${CUCASCADE_PUBLIC_LINK_LIBS}
+                            cucascade_topology_discovery_static)
```
pentschev, line 181: `libcucascade` links to `libcucascade_topology_discovery`, so this shouldn't break any existing use cases.

pentschev (approved):
Because this is blocking cuDF, I'll merge it as soon as CI passes. I briefly discussed this with @felipeblazing and he seems ok with that.

test/CMakeLists.txt
```diff
@@ -62,7 +61,28 @@ target_include_directories(
...
-                                              CUDA::cudart)
+                                              CUDA::cudart_static)
 
 # Register tests with CTest
 add_test(NAME cucascade_tests COMMAND cucascade_tests)
+
+# Create topology discovery test executable
+add_executable(
+  cucascade_topology_discovery_tests memory/test_topology_discovery.cpp
+                                     unittest_topology_discovery.cpp)
```
vyasr, line 72: FWIW you could also just put the definition of `main` in `memory/test_toplogy_discovery.cpp`, but this is fine too.

CMakeLists.txt
```diff
@@ -161,12 +190,29 @@ endif()
...
+
+  target_include_directories(cucascade_topology_discovery_shared
+                             PUBLIC ${CUCASCADE_PUBLIC_INCLUDE_DIRS})
+  target_compile_features(cucascade_topology_discovery_shared PUBLIC cxx_std_20)
+
+  set_target_properties(
+    cucascade_topology_discovery_shared
+    PROPERTIES OUTPUT_NAME cucascade_topology_discovery
+               VERSION ${PROJECT_VERSION}
+               SOVERSION ${PROJECT_VERSION_MAJOR})
```
vyasr, line 206: I suggest that we drop the SOVERSION from both this and any other cucascade libraries. AFAIK there are no ABI stability guarantees being made by cucascade yet.

CMakeLists.txt
vyasr: Ultimately I don't know if it's preferable to link the topology discovery library or the object library into cucascade. Probably fine to stick with the status quo for now.

vyasr (commented):
Some comments for the follow-up, but this looks like a good implementation of my proposal. Thanks Kyle!

PR #129, opened 2026-05-14
Construct `io_worker` thread after all other fields

KyleFromNVIDIA:
When the thread was constructed first, it was potentially started before other fields had a chance to initialize, resulting in strange segfaults and other errors when it encountered invalid fields. In particular, two failure modes were observed:

1. The thread started before the mutex had a chance to initialize, and when it attempted to lock the mutex, an error was thrown.
2. The thread attempted to fetch the pending work function before it was initialized, and thus attempted to call a function in invalid memory, resulting in a segfault.

Construct the thread after all other fields to ensure that they are always valid when the thread starts.

### DIFF
--- a/src/data/pipeline_io_backend.cpp
+++ b/src/data/pipeline_io_backend.cpp
@@ -106,13 +106,15 @@ class io_worker {
     }
   }
 
-  std::thread _thread;
   std::mutex _mutex;
   std::condition_variable _cv;
   std::function<void()> _pending_work;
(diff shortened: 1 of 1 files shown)


wence- (approved):

KyleFromNVIDIA:
I've re-run the entire test suite inside a debugger many times since this change and have not observed any more failures.

pentschev (approved):
Thanks Kyle, indeed looks like a proper fix.

PR #134, opened 2026-05-18
Remove data converter registry

aminaramoon:
remove data registry class. make APIs simpler, automatic conversion from idata_representaiton to other registred types.

### DIFF
--- a/include/cucascade/data/common.hpp
+++ b/include/cucascade/data/common.hpp
@@ -24,8 +24,12 @@
 #include <concepts>
 #include <cstddef>
 #include <memory>
+#include <typeindex>
+#include <typeinfo>
 
 namespace cucascade {
+
+class representation_converter_registry;  // forward declaration for template methods below
 
 /**
  * @brief Interface representing a data representation residing in a specific memory tier.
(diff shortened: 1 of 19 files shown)


src/data/representation_converter.cpp
```diff
@@ -115,22 +116,30 @@ std::unique_ptr<idata_representation> representation_converter_registry::convert
...
-  std::lock_guard<std::mutex> lock(_mutex);
+  std::unique_lock<std::shared_mutex> lock(_mutex);
   _converters.clear();
 }
 
+representation_converter_registry& representation_converter_registry::instance()
+{
+  static representation_converter_registry registry;
+  static std::once_flag init_flag;
+  std::call_once(init_flag, [&] { register_builtin_converters(registry); });
```
mbrobbel, line 139: This `call_once` is redundant for static. We could just use something like `static registry = [] { ...; return r; }()` and drop the `init_flag`.

include/cucascade/data/common.hpp
```diff
@@ -43,8 +47,11 @@ class idata_representation {
    * @brief Construct a new idata_representation object
    *
    * @param memory_space The memory space where the data resides
+   * @param type_idx     The std::type_index of the most-derived class — subclasses must pass
+   *                     `typeid(SubclassName)`. Used as the source key for converter lookup.
```
mbrobbel, line 51: Maybe we should enforce this to prevent runtime issues?

PR #136, opened 2026-05-26
Use pre-pinned HOST pool for cross-GPU host-staged copies

felipeblazing:
When peer DMA is broken (e.g. consumer Intel chipsets), the representation_converter falls back to host-staged copies for cross-GPU transfers. Previously this allocated fresh pinned memory via cudaHostAlloc on every transfer, which is slow and causes significant overhead at scale (SF100+).

Replace the per-transfer cudaHostAlloc with block borrowing from the existing fixed_size_host_memory_resource pool that memory_space already allocates at startup. The staging copy now chunks across the pool's fixed block size (1 MB default), matching the pattern used by convert_gpu_to_host.

Changes:
- Add host pool registry (register/unregister/find_host_pool) in memory/common.{hpp,cpp} so cross-tier code can look up the pre-pinned pool by NUMA ID with cross-NUMA fallback
- memory_space HOST constructor registers its pool; destructor unregisters it
- representation_converter's alloc_and_peer_copy_async uses allocate_multiple_blocks from the pool instead of cudaHostAlloc
- Fallback to one-shot pinned allocation preserved for standalone test builds without a memory_reservation_manager

### DIFF
--- a/include/cucascade/memory/common.hpp
+++ b/include/cucascade/memory/common.hpp
@@ -204,6 +204,33 @@ make_default_host_memory_resource(int device_id, std::size_t capacity, bool make
 
 DeviceMemoryResourceFactoryFn make_default_allocator_for_tier(Tier tier);
 
+// Forward declaration — fixed_size_host_memory_resource.hpp is heavy.
+class fixed_size_host_memory_resource;
+
+/**
+ * @brief Register a HOST-tier `fixed_size_host_memory_resource` so cross-tier
+ * code paths (notably the peer-DMA-broken host-staging branch in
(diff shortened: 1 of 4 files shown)


felipeblazing:
/ok to test [cd712bf](https://github.com/NVIDIA/cuCascade/pull/136/commits/cd712bf52a2f88a8d2eb357ae27ab2f3165e03f4)

felipeblazing:
/ok to test [fad5e93](https://github.com/NVIDIA/cuCascade/pull/136/commits/fad5e9309256f1c53edecbd31dc89920e752387e)

aminaramoon (approved):

PR #138, opened 2026-05-28
Make NVML initialization fork-aware in topology discovery

pentschev:
https://github.com/NVIDIA/cuCascade/pull/115 introduced a change that `topology_discovery::discover()` caches the result of `nvmlInit_v2()` in a process-local static to avoid the driver issue caused by repeated `nvmlInit_v2()`/`nvmlShutdown()` cycles. However, after `fork()`, the child inherits that cached success result even though NVML state is not safely inherited. Subsequent NVML calls in the child can fail, causing topology discovery to report no GPUs.

The fix keeps the no-`nvmlShutdown()` behavior, but keys the cached NVML initialization result by PID. If discovery runs in a forked child, the PID change is detected and `nvmlInit_v2()` is called again in that process before issuing NVML queries.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -24,7 +24,7 @@ unzip = "*"
 # CUDA 13
 [feature.cuda-13]
 system-requirements = { cuda = "13" }
-dependencies = { "cuda-version" = "13.*" }
+dependencies = { "cuda-version" = "13.2.*" }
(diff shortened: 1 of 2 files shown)


mbrobbel (approved):

felipeblazing (approved):

pentschev:
Thanks for the reviews!

PR #141, opened 2026-06-02
Allow topology-only builds at configure time

vyasr:
The topology split in https://github.com/NVIDIA/cuCascade/pull/126 prevented the build of non-topology components of cuCascade when they aren't needed, but it still configures assuming everything is built, which means that it tries to find rmm and cudf. That can still lead to more circular dependency issues depending on exactly how builds are done, so we need a heavier hammer to avoid that. This PR implements a CMake flag to turn off every part of the non-topology build when needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,8 @@ option(CUCASCADE_BUILD_BENCHMARKS "Build the benchmark suite" ON)
 option(CUCASCADE_BUILD_SHARED_LIBS "Build shared library" ON)
 option(CUCASCADE_BUILD_STATIC_LIBS "Build static library" ON)
 option(CUCASCADE_NVTX "Enable NVTX range annotations" OFF)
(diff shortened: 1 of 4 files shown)


pentschev (approved):
Changes being tested in https://github.com/rapidsai/rapidsmpf/pull/1080. Once CI passes there we can merge this.

vyasr:
Converting to draft. This was a workaround that I thought we might need as part of https://github.com/rapidsai/cudf/pull/22747, but at least for the moment we seem to be avoiding any issues in the jobs that do involve refetching cuCascade. I'll keep this in draft for now just in case an edge case pops up while we're moving through the rapidsmpf->cudf transition.

vyasr:
I think we are going to need this to be able to merge https://github.com/rapidsai/cudf/pull/22786 safely.

bdice (approved):

mbrobbel (approved):

PR #145, opened 2026-06-09
Export numa via a bundled FindNuma module instead of an absolute path

### LINKED ISSUE #118: Exported cmake targets embed absolute path to libnuma, breaking cross-distro consumption
This issue was discovered by trying to build rapidsmpf from source in a cudf devcontainer, which pulls in cuCascade. I'm not a cmake expert, but posting Claude's summary of the issue

___

The exported cuCascadeTargets.cmake embeds an absolute path to libnuma.so (e.g. /usr/lib64/libnuma.so), which breaks consumers on systems where libnuma is installed at a different path.

For example, when the cuCascade cmake package is shipped inside a pip wheel built on a manylinux (RHEL-based) system, the exported targets reference /usr/lib64/libnuma.so. On Ubuntu arm64, libnuma lives at /usr/lib/aarch64-linux-gnu/libnuma.so, causing ninja to fail:

```
ninja: error: '/usr/lib64/libnuma.so', needed by 'librapidsmpf.so', missing and no known rule to make it
```

(Human note: Seen in https://github.com/rapidsai/cudf/actions/runs/25450202151/job/74665896167?pr=22281#step:9:14560)

The following line resolves libnuma to an absolute path
https://github.com/NVIDIA/cuCascade/blob/73d00c4f8b59ac478b73236968881f9d3130060f/CMakeLists.txt#L54

That absolute path is then linked as a public dependency

https://github.com/NVIDIA/cuCascade/blob/73d00c4f8b59ac478b73236968881f9d3130060f/CMakeLists.txt#L125

When CMake exports the install targets, ${NUMA_LIB} (e.g. /usr/lib64/libnuma.so) is serialized into cuCascadeTargets.cmake as a literal path.

### Suggested fix

Link with the plain library name instead of the absolute path so the exported targets use -lnuma rather than a hardcoded path:

```
set(CUCASCADE_PUBLIC_LINK_LIBS rmm::rmm cudf::cudf CUDA::cudart
    Threads::Threads numa)
```

The existing `find_library(NUMA_LIB numa REQUIRED)` can remain as a build-time check that the library exists.

Then add a find_library in cuCascadeConfig.cmake.in so consumers also verify libnuma is available:

```
find_library(NUMA_LIB numa REQUIRED)
```

mbrobbel:
## Summary

Fixes #118.

`find_library(NUMA_LIB numa REQUIRED)` resolves `numa` to an absolute path (e.g. `/usr/lib64/libnuma.so`). That path was passed as a `PUBLIC` link dependency via `${NUMA_LIB}`, so CMake serialized it verbatim into the exported `cuCascadeTargets.cmake`. Consumers on systems where libnuma lives at a different path then fail to link — for example a manylinux-built wheel referencing `/usr/lib64/libnuma.so` consumed on Ubuntu arm64 (where it is at `/usr/lib/aarch64-linux-gnu/libnuma.so`):

```
ninja: error: '/usr/lib64/libnuma.so', needed by 'librapidsmpf.so', missing and no known rule to make it
```

## Changes

Rather than embed a library path in the export, this introduces a proper find module that produces an imported target, bundled with the package so the same find logic runs on both sides (per review feedback).

- **`cmake/Modules/FindNuma.cmake`** (new): locates `numa.h` and the `numa` library, validates via `find_package_handle_standard_args`, and produces a `Numa::Numa` imported target carrying both the include directory and the library location. This also propagates the `numa.h` include path, which a bare library reference did not.
- **`CMakeLists.txt`**: use `find_package(Numa REQUIRED)` and link the `Numa::Numa` target instead of a raw path; install `FindNuma.cmake` into the package's `cmake/cuCascade/Modules` dir (skipped for topology-only builds, which don't depend on libnuma).
- **`cmake/cuCascadeConfig.cmake.in`**: prepend the bundled `Modules` dir to `CMAKE_MODULE_PATH` and rerun `find_dependency(Numa REQUIRED)`, so consumers reconstruct `Numa::Numa` against their own libnuma — no path baked into the exported targets, and a clear configure-time error if libnuma is missing.

## Verification

Built, installed to a temp prefix, and configured a standalone consumer project against the install:

- Exported `cuCascadeTargets.cmake` references the target, not a path:
  ```
  INTERFACE_LINK_LIBRARIES "rmm::rmm;cudf::cudf;CUDA::cudart_static;Threads::Threads;Numa::Numa;..."
  ```
- `FindNuma.cmake` is installed under `lib/cmake/cuCascade/Modules/`.
- A consumer running `find_package(cuCascade)` reconstructs `Numa::Numa`, resolving libnuma at its own system path.

---

🤖 This PR was authored by an AI agent (Claude Code), under human review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,8 +64,10 @@ if(NOT CUCASCADE_TOPOLOGY_ONLY)
   find_package(cudf REQUIRED CONFIG)
 
   # Find numa (provided by numactl-devel or libnuma-dev depending on the package
-  # manager)
-  find_library(NUMA_LIB numa REQUIRED)
+  # manager). Uses the bundled FindNuma module, which produces the Numa::Numa
+  # imported target. The same module is installed with the package so consumers
+  # resolve libnuma on their own system rather than inheriting a hardcoded path.
+  find_package(Numa REQUIRED)
 endif()
 
 # Set CUDA architectures
@@ -151,7 +153,7 @@ set(CUCASCADE_PUBLIC_INCLUDE_DIRS
 
 if(NOT CUCASCADE_TOPOLOGY_ONLY)
   set(CUCASCADE_PUBLIC_LINK_LIBS rmm::rmm cudf::cudf CUDA::cudart_static
-                                 Threads::Threads ${NUMA_LIB})
+                                 Threads::Threads Numa::Numa)
 
   # Set include directories for the object library
   target_include_directories(cucascade_objects
@@ -358,6 +360,14 @@ write_basic_package_version_file(
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/cuCascadeConfig.cmake
               ${CMAKE_CURRENT_BINARY_DIR}/cuCascadeConfigVersion.cmake
         DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/cuCascade)
+
+# Bundle the FindNuma module so consumers can rerun the same find logic (the
+# full library links the Numa::Numa target it produces). Not needed for
+# topology-only builds, which do not depend on libnuma.
+if(NOT CUCASCADE_TOPOLOGY_ONLY)
+  install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules/FindNuma.cmake
+          DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/cuCascade/Modules)
+endif()
 
 # =============================================================================
 # Tests
--- a/cmake/Modules/FindNuma.cmake
+++ b/cmake/Modules/FindNuma.cmake
@@ -50,8 +50,8 @@ find_path(Numa_INCLUDE_DIR NAMES numa.h)
 find_library(Numa_LIBRARY NAMES numa)
 
 include(FindPackageHandleStandardArgs)
-find_package_handle_standard_args(
-  Numa REQUIRED_VARS Numa_LIBRARY Numa_INCLUDE_DIR)
+find_package_handle_standard_args(Numa REQUIRED_VARS Numa_LIBRARY
+                                                     Numa_INCLUDE_DIR)
 
 if(Numa_FOUND)
   set(Numa_LIBRARIES ${Numa_LIBRARY})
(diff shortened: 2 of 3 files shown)


pentschev (commented):
@vyasr would you mind having a look at this? This change came from #141 , and I'm no CMake expert to determine whether this change is correct for all situations we support downstream (like conda-forge/wheels). At the very least `CMakeLists.txt` seems to my inexperienced eyes suspicious because apparently linking by `numa` name works, however, `NUMA_LIB` is supposed to be defined prior to that in the same file.

vyasr (changes requested):
This solution will solve the proximate problem, but it doesn't go far enough. The current solution won't guarantee that the numa.h header is available, and it won't propagate any compiler/linker flags or dependencies. What we ought to do is define a find module that produces a suitable NUMA target that we can link to. Then we bundle that file with our package so that the same find logic can run for consumers of cuCascade. [Pytorch's module is one example](https://github.com/pytorch/pytorch/blob/main/cmake/Modules/FindNuma.cmake) that does a more comprehensive search for components and handles the proper find_package arguments. It doesn't create a target though. [cudf's old cufile module is a good example for that](https://github.com/rapidsai/cudf/blob/branch-24.10/cpp/cmake/Modules/FindcuFile.cmake). cudf also demonstrates how you can use a [rapids-cmake function to ensure that find module gets installed for consumers](https://github.com/rapidsai/cudf/blob/branch-24.10/cpp/cmake/thirdparty/get_cufile.cmake#L23). cuCascade doesn't use rapids-cmake, but [the rapids-cmake function in question](https://github.com/rapidsai/rapids-cmake/blob/74a28545e739bc8341cd002c9dae98e03b28939b/rapids-cmake/export/find_package_file.cmake) can be replicated easily enough, it adds the file in question to [a tracked list](https://github.com/rapidsai/rapids-cmake/blob/74a28545e739bc8341cd002c9dae98e03b28939b/rapids-cmake/export/find_package_file.cmake#L78C76-L78C100) and later [copies them into the build/install tree](https://github.com/rapidsai/rapids-cmake/blob/74a28545e739bc8341cd002c9dae98e03b28939b/rapids-cmake/export/write_dependencies.cmake#L98).

pentschev (commented):
To my inexperienced eyes, this looks sane to me. @vyasr could you have another look?

vyasr (approved):
Looks right to me, thanks!

pentschev (approved):
Thanks @mbrobbel for working on it and @vyasr for reviews.

PR #146, opened 2026-06-10
Implemented ability to rebind_stream

wmalpica:
Currently all deallocations happen on the stream on which they allocations were made. This PR adds an API which allows for changing that stream, so that deallocations can be made to happen on the active stream.

It also changes the exclusive_stream_pool class so that it recycles streams more in a FIFO manner instead of LIFO. This way it cycles through all streams in a round robin fashion


mbrobbel:
/ok to test [0def118](https://github.com/NVIDIA/cuCascade/pull/146/commits/0def11815e35234e2384c7d4dfaf333b9252fd10)

bwyogatama (approved):

PR #149, opened 2026-06-11
Keep LIST offsets INT32 during H2D and disk reconstruction

### LINKED ISSUE #147: H2D/disk reconstruction casts LIST offsets child to INT64, producing malformed cuDF LIST columns
### Summary
When reconstructing a cuDF table during H2D conversion, the offsets child of LIST columns is cast from INT32 to INT64. cuDF LIST columns require INT32 (`cudf::size_type`) offsets, so the resulting column is malformed: list algorithms (which read the offsets as `size_type`) misinterpret it and the sublists collapse.

### Location
`src/data/representation_converter.cpp`:
- `reconstruct_column` (H2D path), LIST branch: lines ~1147–1165 (the INT32→INT64 cast at ~1153–1158)
- `reconstruct_column_from_disk` (disk→GPU path), LIST branch: lines ~1772–1789 (cast at ~1780–1783)

### Root cause
The INT32→INT64 offsets cast is correct for STRING columns (cuDF's large-strings convention uses 64-bit offsets), but it was applied to the LIST branch as well. cuDF's `lists_column_view` and list algorithms treat the offsets child as `size_type` (INT32). `make_lists_column` does not validate or re-cast the offsets type, so the malformed column is constructed silently.

### Expected
For LIST columns, the offsets child should remain INT32 after reconstruction. The INT64 promotion should apply only to STRING offsets.

### Actual
LIST offsets are promoted to INT64; downstream cuDF list operations read the offsets incorrectly and collapse the sublists.

### Impact / workaround
We currently work around it by recasting every LIST column's offsets back to INT32 immediately after the H2D conversion. 

### Environment
- cuCascade at commit `b4abc2d64cc9ade1efe252f69d27cd2300d9c94e`
- libcudf 26.06

mbrobbel:
## Summary

cuDF `LIST` columns require `INT32` (`cudf::size_type`) offsets, but the H2D (`reconstruct_column`) and disk→GPU (`reconstruct_column_from_disk`) reconstruction paths cast the LIST offsets child from `INT32` to `INT64`. That promotion is correct for `STRING` offsets (cuDF's large-strings convention) but malformed for `LIST`: `lists_column_view` and list algorithms read the offsets child as `size_type`, so the sublists collapse. `make_lists_column` does not validate or re-cast the offsets type, so the malformed column was constructed silently.

This builds the `LIST` column directly with `[offsets, values]` children (mirroring the existing `STRUCT` branch) so the offsets stay `INT32`. As a side benefit it skips `make_lists_column`'s `purge_nonempty_nulls` kernel, which would otherwise read offset/child buffers still pending in the batched H2D copies.

The `STRING` branches are unchanged (their `INT64` promotion is correct), and the GPU→GPU peer path already preserved the offsets type.

Closes #147

## Changes

- `src/data/representation_converter.cpp`: direct `cudf::column` construction for the LIST branch in both `reconstruct_column` and `reconstruct_column_from_disk`.
- `test/utils/cudf_test_utils.cpp`: the LIST comparison now requires the offsets child type to match, turning every list round-trip test into a regression guard for this bug.
- `test/data/test_gpu_disk_converters.cpp`, `test/data/test_disk_host_converters.cpp`: new flat and nested (`LIST<LIST<INT32>>`) nullable-list round-trip tests, covering both reconstruction paths.

## Test plan

- Confirmed the strengthened tests **fail** on the pre-fix behavior (temporarily reintroducing the `INT64` cast → `LIST offsets type mismatch: left=3 right=4`) and **pass** with the fix.
- Full suite green: **5643 assertions across 284 test cases** (built/run against the cuda-13 stable env, libcudf 26.06 — matches the reported environment).


mbrobbel:
/ok to test [a6f0bad](https://github.com/NVIDIA/cuCascade/pull/149/commits/a6f0badc15344594906cfdcf3400a6da8d6bf062)

Jedi18 (approved):
nit: should we add a test for nested lists?

felipeblazing (approved):

PR #151, opened 2026-06-17
allow release of fixed size buffers in cucascade

aminaramoon:
Allow release of chunks.

this is needed when we want to allocate a large number of chunks, but we can only release them in small parts. so we can release the buffers, form new blocks and release them back to cucascade. 

---

use case: fs cache built on top of cucascade.
where we allocate 600 MBs for a file, then some chunks in that file are requeted again in a different request, so we keep them around but return the left overs to cucascade/


mbrobbel:
/ok to test [f46647d](https://github.com/NVIDIA/cuCascade/pull/151/commits/f46647d5fa1025200671eed028123a526c6e2188)

felipeblazing (approved):

PR #154, opened 2026-07-01
Standardize `data_batch` on `std::shared_ptr`

dhruv9vats:
Standardizes `data_batch` on `std::shared_ptr` ownership and removes the repository pointer-type templates.

`data_batch` inherits from `std::enable_shared_from_this` and its transition APIs call `shared_from_this()`. This _requires_ every instance to be managed by a `std::shared_ptr`; constructing or storing a batch with unique ownership would violate that requirement and will result in `std::bad_weak_ptr`.

Since shared ownership is already a prerequisite, supporting `std::unique_ptr<data_batch>` adds complexity without providing a valid usage model. This change enforces factory-based construction and simplifies repositories and managers to support only `std::shared_ptr<data_batch>`.


include/cucascade/data/data_repository_manager.hpp
```diff
@@ -127,17 +122,20 @@ class data_repository_manager {
...
-   * For unique_ptr: This method requires only one operator (single owner semantics).
+   * The shared batch pointer is copied to each repository.
    *
    * @param batch The data_batch smart pointer to add
    * @param ops The operator IDs and ports whose repositories will receive this batch
    *
    * @note Thread-safe operation
    */
-  void add_data_batch(PtrType batch, std::vector<std::pair<size_t, std::string_view>> ops)
+  void add_data_batch(const std::shared_ptr<data_batch>& batch,
```
mbrobbel: This uses `const std::shared_ptr<data_batch>&` vs `std::shared_ptr<data_batch>` in `data_repository`, which means ownership doesn't always transfer here. We should match them or document the lifetime semantics.

dhruv9vats: [Use value args for consistency](https://github.com/NVIDIA/cuCascade/pull/154/commits/cd9db6d7602b08fcb7c8b89da5d9c45488f117ba)

mbrobbel (approved):

wmalpica (approved):

PR #156, opened 2026-07-03
Add `data_batch_probe`

dhruv9vats:
Add a virtual `idata_batch_probe` that consumers can override to inspect or probe `data_batch` behavior.

### DIFF
--- a/include/cucascade/data/data_batch.hpp
+++ b/include/cucascade/data/data_batch.hpp
@@ -52,6 +52,7 @@ enum class batch_state { idle, read_only, mutable_locked };
 // Forward declarations -- required before data_batch because it friends them.
 class read_only_data_batch;
 class mutable_data_batch;
+class idata_batch_probe;
 
 /**
  * @brief Core data batch type representing the "idle" (unlocked) state.
@@ -80,8 +81,10 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
    * @return A shared pointer owning the new batch.
    * @throws std::runtime_error if data is null.
    */
-  static std::shared_ptr<data_batch> make(uint64_t batch_id,
-                                          std::unique_ptr<idata_representation> data);
+  static std::shared_ptr<data_batch> make(
+    uint64_t batch_id,
+    std::unique_ptr<idata_representation> data,
+    std::unique_ptr<idata_batch_probe> probe = std::make_unique<idata_batch_probe>());
 
   ~data_batch() = default;
 
@@ -228,7 +231,9 @@ class data_batch : public std::enable_shared_from_this<data_batch> {
   [[nodiscard]] static read_only_data_batch mutable_to_readonly(mutable_data_batch&& accessor);
 
  private:
-  data_batch(uint64_t batch_id, std::unique_ptr<idata_representation> data);
+  data_batch(uint64_t batch_id,
+             std::unique_ptr<idata_representation> data,
+             std::unique_ptr<idata_batch_probe> probe);
 
   /**
    * @brief Get the memory tier of the held data.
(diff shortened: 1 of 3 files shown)


dhruv9vats:
/ok to test [0163f0e](https://github.com/NVIDIA/cuCascade/pull/156/commits/0163f0ee9a545a93f1bed4d2af618d7c12f14ae1)

src/data/data_batch.cpp
```diff
@@ -68,7 +81,12 @@ memory::memory_space* data_batch::get_memory_space() const
   return &_data->get_memory_space();
 }
 
-void data_batch::set_data(std::unique_ptr<idata_representation> data) { _data = std::move(data); }
+void data_batch::set_data(std::unique_ptr<idata_representation> data)
+{
+  if (_data == nullptr) { throw std::runtime_error("data is null in data_batch::set_data"); }
```
mbrobbel:
```suggestion
  if (data == nullptr) { throw std::runtime_error("data is null in data_batch::set_data"); }
```

include/cucascade/data/data_batch.hpp
```diff
@@ -501,6 +513,41 @@ class mutable_data_batch {
...
+ * return quickly, as they are called in a thread-safe manner, and will block other
+ * mutating changes while they execute. This interface is primarily intended for
+ * bookkeeping purposes. Default impl is no-op
+ */
+class idata_batch_probe {
+ public:
+  idata_batch_probe()          = default;
+  virtual ~idata_batch_probe() = default;
+
+  virtual void created([[maybe_unused]] const uint64_t batch_id,
```
mbrobbel, line 538: These callbacks can throw, but several are invoked from destructors and `noexcept` paths, causing `std::terminate`. Please make the callback contract `noexcept` or contain exceptions at call sites.

src/data/data_batch.cpp
```diff
@@ -17,25 +17,37 @@
...
+  _probe->created(get_batch_id(), *get_data());
 }
 
 uint64_t data_batch::get_batch_id() const { return _batch_id; }
 
-void data_batch::subscribe() { _subscriber_count.fetch_add(1, std::memory_order_relaxed); }
+void data_batch::subscribe()
+{
+  const size_t prev = _subscriber_count.fetch_add(1, std::memory_order_relaxed);
+  _probe->subscriber_count_changed(prev + 1);
```
mbrobbel: Concurrent updates can deliver count callbacks out of order (e.g. `2` before `1`). Please serialize notifications or use ordered/delta-based events. This also affects reader-count callbacks.

dhruv9vats: [Guard probe notifications with a mutex to avoid races](https://github.com/NVIDIA/cuCascade/pull/156/commits/597ee19784ebd5774e4da4afdcd0af851b8dd349)

include/cucascade/data/data_batch.hpp
dhruv9vats: Updated the contract to be `noexcept`. Thank you.

src/data/data_batch.cpp
mbrobbel: Conversion and data replacement callbacks still bypass `_probe_mutex`. `subscribe()` can enter the same probe concurrently with `convert_to()` or `set_data()`.

Also, callbacks now execute while holding a non-recursive mutex, so we should add a note about this to avoid a deadlock when callbacks call the same function again that tries to lock the same mutex again.

dhruv9vats: [Only keep callbacks already under unique_lock](https://github.com/NVIDIA/cuCascade/pull/156/commits/f2e12c290a0b58a6ba84bd4577affca536755294)

mbrobbel (approved):

PR #158, opened 2026-07-08
Stream-order slab reuse in small_pinned_host_memory_resource

wmalpica:
This PR is to fix the query validation error https://github.com/sirius-db/sirius/issues/1060
The root bug is in cudf https://github.com/rapidsai/cudf/issues/23178
This PR makes things work safely for now.

**Stream-order slab reuse in small_pinned_host_memory_resource**

The slab pool recycled freed pinned blocks immediately, ignoring the stream: deallocate pushed the slab straight onto the free list and allocate handed it back to the next caller with no ordering. When a consumer had an async H2D copy still reading a slab (e.g. cuDF's Parquet stats-filter min/max buffers, whose host_column::to_device copies async and doesn't synchronize before the source is freed), another task on a different stream could reuse that slab and overwrite it mid-copy â corrupting the data. In Sirius this surfaced as an intermittent (~1/800) filtered scan returning 0 rows (all row groups wrongly pruned) â wrong query results.

Change: make slab reuse stream-ordered via CUDA events.
- deallocate(stream, â¦) records a timing-disabled event on the freeing stream (capturing any in-flight use of the slab) and stores it alongside the slab on the free list.
- allocate(stream, â¦) issues cudaStreamWaitEvent before returning a recycled slab to a different stream, so the reusing stream can't touch the slab until the prior work completes; the event is then returned to a small internal pool for reuse.
- Freshly-carved slabs carry no event; events are best-effort (a creation/record failure just falls back to unordered reuse) and are destroyed in the destructor.

Buffers stay pinned (required â cuDF reads some of these directly from GPU kernels, so making them pageable would cause cudaErrorIllegalAddress); this only adds the missing cross-stream ordering.

Files: include/cucascade/memory/small_pinned_host_memory_resource.hpp, src/memory/small_pinned_host_memory_resource.cpp.

### DIFF
--- a/include/cucascade/memory/small_pinned_host_memory_resource.hpp
+++ b/include/cucascade/memory/small_pinned_host_memory_resource.hpp
@@ -21,6 +21,7 @@
 
 #include <cuda/memory_resource>
 #include <cuda/stream_ref>
+#include <cuda_runtime_api.h>
 
 #include <array>
 #include <cstddef>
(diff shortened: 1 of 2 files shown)


mbrobbel:
/ok to test [b7f16cf](https://github.com/NVIDIA/cuCascade/pull/158/commits/b7f16cf2199acb9c60136cdd2f192b01252d71e1)

bwyogatama (approved):

PR #163, opened 2026-07-16
adds ability to set_num_partitions in a data_repository

wmalpica:
This PR is a companion to https://github.com/sirius-db/sirius/pull/1194

All it does is add a new API which allows for setting the number of partitions on a data_repository.
Before we were implicitly doing that by adding data_batches into a partition index, so it would resize automatically. That functionality is still there, but this allows you to explicitly set that and it will resize its internal containers

### DIFF
--- a/include/cucascade/data/data_repository.hpp
+++ b/include/cucascade/data/data_repository.hpp
@@ -23,6 +23,8 @@
 #include <memory>
 #include <mutex>
 #include <optional>
+#include <stdexcept>
+#include <string>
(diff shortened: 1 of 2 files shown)


mbrobbel:
/ok to test [4795f6b](https://github.com/NVIDIA/cuCascade/pull/163/commits/4795f6b1d5797f7e9d1bf856d611a64e9e6a05c6)

mbrobbel (approved):

PR #170, opened 2026-07-28
Upgrade Catch2 from v2.13.10 to v3.8.1

aminaramoon:
## Summary

- Bumps Catch2 from v2.13.10 to v3.8.1, eliminating the CMake deprecation warning emitted by v2's `cmake_minimum_required(VERSION 2.8.8)` under CMake 3.27+
- Removes manual `single_include/` include paths from all three test targets (v3 exposes headers via its CMake target)
- Replaces `<catch2/catch.hpp>` with `<catch2/catch_all.hpp>` across all 17 test files
- Renames `Catch::TestEventListenerBase` → `Catch::EventListenerBase` in `unittest.cpp` (class was renamed in v3)

## Test plan

- [x] `pixi run build` — clean build, no deprecation warnings
- [x] `pixi run test` — all 3 test suites pass (100%, 13.34s)

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -20,7 +20,7 @@ include(FetchContent)
 FetchContent_Declare(
   Catch2
   GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-  GIT_TAG v2.13.10)
+  GIT_TAG v3.8.1)
 FetchContent_MakeAvailable(Catch2)
 
 if(NOT CUCASCADE_TOPOLOGY_ONLY)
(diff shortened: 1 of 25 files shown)


felipeblazing (approved):

dhruv9vats (approved):

test/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ include(FetchContent)
 FetchContent_Declare(
   Catch2
   GIT_REPOSITORY https://github.com/catchorg/Catch2.git
-  GIT_TAG v2.13.10)
+  GIT_TAG v3.8.1)
```
mike-wendt:
```suggestion
  GIT_TAG v3.8.1
  GIT_SHALLOW TRUE)
```

Prevents cloning the whole history of the repo

test/unittest_topology_discovery.cpp
```diff
@@ -17,7 +17,7 @@
 
 #define CATCH_CONFIG_RUNNER
```
mike-wendt:
```suggestion
```

Suggested by Claude, to remove this define from v2 that is a no-op in v3

mike-wendt (approved):

test/unittest.cpp
mike-wendt: I'd remove this as well, suggestion isn't picking it up

PR #172, opened 2026-07-29
feat(io): S3 LIST + footer probe + kvikIO backend, and io-layer fixes

aminaramoon:
Adds three capabilities to the `io` datasource layer, fixes several latent defects found while auditing it, and gives the layer its first test coverage.

## Bug fixes

- **`io_request`: `copy_async` validates its host source before forming the pointer.** A null `host_buffer`, or an out-of-range `[src_off, src_off + size)`, previously produced UB (`nullptr + offset`) or a wild in-range pointer that the near-null check could not catch. Now returns `cudaErrorInvalidValue`. The existing asserts compile out in release, so the check has to happen before the pointer is formed.
- **`prefetching_cache`: mark the *request* state load-failed on the `prefetch_loop` cancel/stop path.** Only the chunk states were being marked, leaving the request-level state inconsistent on cancellation.
- **Four empty `catch` blocks were silently swallowing exceptions** — the `rest_reactor` and `uring_reactor` worker loops, and two in `templated_ioctx::shutdown`. Restoring the log call sites fills them. `CUCASCADE_LOG_*` are no-ops today, so this costs nothing at runtime but keeps the diagnostics in place for when logging is wired up.

## Behavioral fixes

- **`rest_reactor::preferred_prefetching_stage`: `opportunistic` → `just_in_time`.** `prefetching_cache` branches on exactly this value, so S3 reads were being eagerly prefetched into cache buffers rather than read straight into the caller's buffer. Network round-trips are high-latency; read ahead on demand rather than prefilling the working set.
- **`uri_parser`: S3 object keys are literal bytes.** `%`, `?` and `#` are ordinary key bytes, matching AWS CLI semantics, so `s3://b/a%20b` now opens the key `a%20b`. Previously the key was percent-decoded and split on `?`, which opened the wrong object.

## New capability

- **S3 ListObjectsV2** — a hand-rolled, fail-closed list parser (no XML dependency); `presign_url(extra_canonical_query)` so caller query params participate in the signature (required for S3 to accept a presigned LIST); `authorize_list` on the authorizer interface and both SigV4 implementations; an `X-Amz-*` injection guard so callers cannot smuggle or override signing params; `rest_reactor::list_page`; and `rest_ioctx::list_objects_paged` / `list_objects` / `list_max_matches`. Pagination is guarded against non-conforming backends: truncated-without-token, truncated-but-empty, and non-advancing continuation token all throw rather than loop or silently truncate.
- **Parquet footer probe** — `open_hint` plus `create_io_object(path, hint)`; a suffix-range GET that resolves the object size *and* stashes its trailing bytes in one round-trip; and a stash-hit fast path in `rest_reactor::host_read` so the footer reads that follow are served from memory instead of costing extra round-trips. Falls back to a plain HEAD on any unusable response (200 full body, 416, missing/unsatisfied `Content-Range`).
- **Known-size open** — `create_io_object(path, known_size)`, so a size already learned from a LIST response builds the io_object with zero network: no HEAD, no probe.
- **`kvikio_context`** — a local-file fallback backend built directly on `kvikio::FileHandle` rather than `cudf::io::datasource`, which keeps the io library cudf-free. Registered as a catch-all that `lookup_path` defers behind the explicit uring/rest backends, so `s3://` never resolves to it and a local file still routes to uring first.
- **`kvikio_config`** — optional-per-field tunables: `nthreads`, `task_size`, `gds_threshold`, `bounce_buffer_size`, O_DIRECT reads and overread, per-block-device pools, and compat mode. An unset field means "leave kvikIO's own default alone", so the `KVIKIO_*` environment variables keep working as the outer default and this is an explicit in-process override layered on top. Every field except `compat_mode` maps to a setter on kvikIO's process-global `defaults` singleton and is documented as such — last context constructed wins. `compat_mode` is the exception: it rides the `FileHandle` constructor, so it scopes to files this ioctx opens and mutates nothing global.

## Cleanup

- `metadata_store::get_metadata(cache_key)` overload, for callers that know the path but have not built an io_object yet.
- Dropped `object_store_config::s3_use_async_backend` — no consumers, and it referenced types that do not exist.
- Renamed `io_context_registry::lookup` → `lookup_path`. It is passed a full path, not a bare scheme (the checkers parse the URI / stat the filesystem themselves), and the docs claimed otherwise.

## Tests

New `cucascade_io_tests` target — **732 assertions across 71 cases**, covering `uri_parser`, sigv4, the list parser, both authorizers, static credentials and `kvikio_config`. **The io layer had no tests at all before this.**

Full suite: **6,330 assertions / 343 cases, 0 failures.**

| Binary | Result |
|---|---|
| `cucascade_tests` | 3225 / 89 |
| `cucascade_io_tests` (new) | 732 / 71 |
| `cucascade_cudf_tests` | 2326 / 177 |
| `cucascade_topology_discovery_tests` | 47 / 6 |

Builds clean with `CUCASCADE_WARNINGS_AS_ERRORS=ON`; `pre-commit run -a` passes.

## Not verified

The two S3 benchmarks are gated behind `CUCASCADE_BUILD_S3_BENCHMARK`, which requires aws-sdk-cpp (present only in the `s3-bench` pixi environment), so **they were not compiled**. By inspection they use only the single-argument `open_datasource` / `open_io_object` overloads — unchanged, since the new behavior was added as overloads rather than signature changes — and nothing switches over `io_context_type`, so the new `kvikio` enumerator cannot trip `-Werror=switch`.

## Follow-up

REST perf instrumentation (per-chunk timings, queue wait, TTFB, retry/terminal counters, pool aggregation) is deliberately not part of this PR. The control-plane paths added here leave room for it and it layers on cleanly; the existing S3 benchmarks are its natural consumer.

### DIFF
--- a/include/cucascade/cudf/datasource.hpp
+++ b/include/cucascade/cudf/datasource.hpp
@@ -159,4 +159,19 @@ class datasource : public cudf::io::datasource {
 [[nodiscard]] std::unique_ptr<datasource> open_datasource(std::shared_ptr<ioctx> io_ctx,
                                                           std::string path);
 
+/// As above, forwarding @p hint to the backend's io_object resolution so it can,
+/// e.g., prefetch a parquet footer in the same round-trip as the size
+/// (@c open_hint::parquet_footer_probe).  Backends that cannot act on the hint
+/// fall back to the plain open.
+[[nodiscard]] std::unique_ptr<datasource> open_datasource(std::shared_ptr<ioctx> io_ctx,
+                                                          std::string path,
+                                                          open_hint hint);
+
+/// As above, with the object's size already known (e.g. from an S3
+/// ListObjectsV2 response), so a backend that can act on it skips its size
+/// discovery entirely (no HEAD for object stores).
+[[nodiscard]] std::unique_ptr<datasource> open_datasource(std::shared_ptr<ioctx> io_ctx,
+                                                          std::string path,
+                                                          std::uint64_t known_size);
+
 }  // namespace cucascade::io
--- a/include/cucascade/io/rest/config.hpp
+++ b/include/cucascade/io/rest/config.hpp
@@ -18,6 +18,8 @@
 
 #pragma once
 
+#include <cucascade/io/rest/s3/list_parser.hpp>
+
 #include <chrono>
 #include <cstddef>
 
@@ -84,6 +86,27 @@ struct config {
   std::chrono::milliseconds retry_backoff_base{50};
   std::chrono::milliseconds retry_jitter{50};
   bool honor_retry_after{true};
+
+  /// Suffix-range window (bytes) for the parquet footer probe
+  /// (@c open_hint::parquet_footer_probe): one `Range: bytes=-N` GET resolves the
+  /// object size and stashes its last N bytes, so cuDF's trailer/footer reads are
+  /// served locally.  Tradeoff — a parquet footer is ~0.037% of the file (SF1
+  /// lineitem 207 MB -> 78 KiB, SF10 2.2 GB -> 771 KiB): N must cover the footer,
+  /// else the probe wastes the suffix and re-GETs the footer body (worse than a
+  /// plain HEAD), so err large; the over-read when N exceeds the footer is a
+  /// one-time bind transfer (~10 ms on a high-bandwidth link).  The 512 KiB
+  /// default covers files up to ~1.4 GB in one GET (the common range); raise it
+  /// for multi-GB single files, lower it for many-tiny-file / low-bandwidth
+  /// workloads.
+  std::size_t footer_probe_bytes{512UL << 10};  // 512 KiB
+
+  /// S3 LIST / glob safety caps (both throw "narrow the glob prefix", never
+  /// truncate).  @c list_max_matches bounds the files a glob keeps / a
+  /// whole-listing accumulates (result memory); @c list_max_scanned bounds the
+  /// objects a LIST sweep looks at across pages (time / LIST round-trips).  The
+  /// two axes diverge when a prefix is huge but few keys match, so both exist.
+  std::size_t list_max_matches{s3::default_max_list_objects};     // 100'000
+  std::size_t list_max_scanned{s3::default_max_scanned_objects};  // 1'000'000
 };
 
 }  // namespace cucascade::io::rest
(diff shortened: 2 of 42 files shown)


include/cucascade/cudf/datasource.hpp
```diff
@@ -159,4 +159,19 @@ class datasource : public cudf::io::datasource {
 [[nodiscard]] std::unique_ptr<datasource> open_datasource(std::shared_ptr<ioctx> io_ctx,
                                                           std::string path);
 
+/// As above, forwarding @p hint to the backend's io_object resolution so it can,
```
felipeblazing, line 162: we can probably clean up the AI comments after reviews finish. I don't mind the explanations while reviewing.

include/cucascade/io/rest/config.hpp
```diff
@@ -84,6 +86,27 @@ struct config {
...
+  /// object size and stashes its last N bytes, so cuDF's trailer/footer reads are
+  /// served locally.  Tradeoff — a parquet footer is ~0.037% of the file (SF1
+  /// lineitem 207 MB -> 78 KiB, SF10 2.2 GB -> 771 KiB): N must cover the footer,
+  /// else the probe wastes the suffix and re-GETs the footer body (worse than a
+  /// plain HEAD), so err large; the over-read when N exceeds the footer is a
+  /// one-time bind transfer (~10 ms on a high-bandwidth link).  The 512 KiB
+  /// default covers files up to ~1.4 GB in one GET (the common range); raise it
+  /// for multi-GB single files, lower it for many-tiny-file / low-bandwidth
+  /// workloads.
+  std::size_t footer_probe_bytes{512UL << 10};  // 512 KiB
```
felipeblazing, line 101: If retrieving a larger payload doesn't increase the latency significantly it woyuld be fine to increase this a bit.

include/cucascade/io/rest/mock_authorizer.hpp
```diff
@@ -0,0 +1,139 @@
+/*
```
felipeblazing, line 1: I think this file should be moved into the testing directory

felipeblazing: Or is the idea that other projects are going to want to use this authorizer themselves for testing? If so maybe we can make a new target for things that aren't for production but are meant to be included by users of the library for testing?

src/io/rest/s3/list_parser.cpp
```diff
@@ -0,0 +1,252 @@
...
+#include <stdexcept>
+
+namespace cucascade::io::rest::s3 {
+
+namespace {
+
+// Single-pass unescape of the five predefined XML entities. Unknown sequences
+// (e.g. numeric character references, which S3 does not emit for keys) pass
+// through verbatim.
+std::string xml_unescape(std::string_view s)
```
felipeblazing, line 33: It seems risky to handle xml parsing ourselves. Is there a reason we can't just use a library that handles that for us?

aminaramoon: I agree, the alternative is boost

felipeblazing (approved):

PR #173, opened 2026-07-29
feat: detect NUMA region capacity and size host spaces from it

aminaramoon:
Topology discovery now enumerates NUMA nodes with their total and free
memory, parsed from /sys/devices/system/node/node<id>/meminfo, and
num_numa_nodes is derived from that list. system_topology_info gains
get_numa_memory_capacity(), get_numa_free_memory(), and
get_total_numa_memory_capacity() lookups.

reservation_manager_configurator gains set_usage_limit_ratio_per_host(),
which sizes each host memory space as a fraction of the capacity of the
NUMA node backing it, mirroring set_usage_limit_ratio_per_gpu(). Absolute
per-host and total-host capacities keep their existing behavior; build()
throws when a fraction is requested and the NUMA capacity is unknown.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/include/cucascade/memory/reservation_manager_configurator.hpp
+++ b/include/cucascade/memory/reservation_manager_configurator.hpp
@@ -120,6 +120,14 @@ class reservation_manager_configurator {
   /// @param bytes Memory capacity per NUMA node in bytes.
   builder_reference& set_per_host_capacity(std::size_t bytes);
 
+  /// @brief set the capacity of each host tier as a fraction of its NUMA node capacity
+  /// @param fraction Fraction of the NUMA node memory capacity to use, in (0.0, 1.0].
+  /// @note Requires NUMA capacities in the topology passed to `build()`; `build()` throws
+  ///       if the capacity of a NUMA node backing a host space is unknown.
+  /// @note The fraction applies per host space. With `use_host_per_gpu()` and several GPUs
+  ///       on the same NUMA node, each space gets that fraction of the shared node.
+  builder_reference& set_usage_limit_ratio_per_host(double fraction);
+
   /// \brief set reservation limit ratio per GPU
   /// @param fraction Fraction of GPU memory capacity to reserve.
   builder_reference& set_downgrade_fractions_per_host(double start, double end);
@@ -171,6 +179,7 @@ class reservation_manager_configurator {
   struct host_info {
     int space_id{-1};
     int numa_id{-1};
+    std::size_t numa_capacity{0};
   };
 
   std::vector<gpu_info> extract_gpu_ids(const system_topology_info& topology) const;
@@ -191,6 +200,16 @@ class reservation_manager_configurator {
         auto size = std::get<std::size_t>(_fraction_or_size_value);
         return static_cast<double>(size) / static_cast<double>(total_size);
       }
+    }
+
+    [[nodiscard]] bool holds_fraction() const
+    {
+      return std::holds_alternative<double>(_fraction_or_size_value);
+    }
+
+    [[nodiscard]] std::size_t get_size() const
+    {
+      return std::get<std::size_t>(_fraction_or_size_value);
     }
 
     [[nodiscard]] std::size_t get_capacity(std::size_t total_size) const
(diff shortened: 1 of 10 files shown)


include/cucascade/memory/reservation_manager_configurator.hpp
```diff
@@ -120,6 +120,14 @@ class reservation_manager_configurator {
   /// @param bytes Memory capacity per NUMA node in bytes.
   builder_reference& set_per_host_capacity(std::size_t bytes);
 
+  /// @brief set the capacity of each host tier as a fraction of its NUMA node capacity
+  /// @param fraction Fraction of the NUMA node memory capacity to use, in (0.0, 1.0].
+  /// @note Requires NUMA capacities in the topology passed to `build()`; `build()` throws
+  ///       if the capacity of a NUMA node backing a host space is unknown.
+  /// @note The fraction applies per host space. With `use_host_per_gpu()` and several GPUs
+  ///       on the same NUMA node, each space gets that fraction of the shared node.
+  builder_reference& set_usage_limit_ratio_per_host(double fraction);
```
felipeblazing:
```suggestion
  builder_reference& set_usage_limit_ratio_per_numa_region(double fraction);
```

include/cucascade/memory/topology_discovery.hpp
```diff
@@ -61,6 +75,51 @@ struct system_topology_info {
...
+  }
+
+  /**
+   * @brief Get the free memory of a NUMA node.
+   *
+   * @param numa_id NUMA node ID to look up.
+   * @return Free memory of the node in bytes; 0 if the node is unknown or its free
+   *         memory could not be determined.
+   */
+  [[nodiscard]] std::size_t get_numa_free_memory(int numa_id) const
```
felipeblazing: it feels like if the user sends in the wrong numa_id this and get_numa_memory_capacity will silently fail

src/memory/reservation_manager_configurator.cpp
```diff
@@ -125,6 +126,14 @@ builder_reference& reservation_manager_configurator::set_per_host_capacity(std::
   return *this;
 }
 
+builder_reference& reservation_manager_configurator::set_usage_limit_ratio_per_host(double fraction)
```
felipeblazing:
```suggestion
builder_reference& reservation_manager_configurator::set_usage_limit_ratio_per_numa_region(double fraction)
```

src/memory/reservation_manager_configurator.cpp
```diff
@@ -313,17 +334,20 @@ reservation_manager_configurator::extract_gpu_ids(const system_topology_info& to
 }
 
 std::vector<reservation_manager_configurator::host_info>
-reservation_manager_configurator::extract_host_ids(
-  const std::vector<gpu_info>& gpus, [[maybe_unused]] const system_topology_info& topology) const
+reservation_manager_configurator::extract_host_ids(const std::vector<gpu_info>& gpus,
```
felipeblazing: I might consider renaming this to numa_region_ids to be clearer.

src/memory/reservation_manager_configurator.cpp
```diff
@@ -225,12 +234,24 @@ std::vector<memory_space_config> reservation_manager_configurator::build(
     configs.emplace_back(config);
   };
 
-  size_t per_host_capacity =
-    (host_infos.size() <= 1)
-      ? _host_capacity
-      : (_is_capacity_per_space ? _host_capacity : _host_capacity / host_infos.size());
+  // Absolute capacities are either per-space or split evenly across spaces; a fraction is
+  // always relative to the capacity of the NUMA node backing the space.
+  auto host_capacity_of = [&](const host_info& info) -> std::size_t {
```
felipeblazing:
```suggestion
  auto numa_region_capacity_of = [&](const host_info& info) -> std::size_t {
```

src/memory/topology_discovery.cpp
```diff
@@ -695,32 +695,93 @@ std::string get_hostname()
...
+/**
+ * @brief Discover NUMA nodes and their memory capacities.
+ *
+ * Scans subdirectories named "node<id>" under /sys/devices/system/node and reads each
+ * node's memory capacity. Returns an empty vector if the directory does not exist or
+ * cannot be iterated.
+ *
+ * @return NUMA node information sorted by node id; empty if unavailable.
+ */
+std::vector<numa_topology_info> discover_numa_nodes()
```
felipeblazing, line 771: Is there a way that we can figure out during numa_node discovery whether or not a numa_node  is actually gpu memory as happens on the dgx station for example where the gpu memory shows up as numa region 1?

felipeblazing (changes requested):
In general I think we should change most of our host nomenclature to numa_region so that this is clear what we are talking about. This would prevent downstream users from getting confused between the entire host capacity and a numa region

felipeblazing (approved):

felipeblazing:
/ok to test [c6bed56](https://github.com/NVIDIA/cuCascade/pull/173/commits/c6bed56dc6146cab77c72a45ab69d93565f51b56)

PR #175, opened 2026-07-30
create a the log hooks for cucascade

aminaramoon:


src/log/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+# Record delivery (vemit), sink publication and the built-in stderr sink. Only
+# the level check and the sink slot stay inline in
+# include/cucascade/log/logging.hpp, so the hot path remains header-only while
+# <format>, <chrono> and <ctime> stay out of every translation unit that logs.
+#
+# Logging therefore requires linking cuCascade, and is unavailable in a
+# topology-only build, which does not produce cucascade_objects.
+if(TARGET cucascade_objects)
+  target_sources(cucascade_objects
+                 PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/logging.cpp)
```
ran-yuan-rui, line 27: seems like we only build logging.cpp but no stderr_sink.cpp?

src/log/stderr_sink.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include <ctime>
+#include <iomanip>
+#include <locale>
+#include <sstream>
+#include <string_view>
+
+namespace cucascade::log {
+inline namespace v1 {
+
+void stderr_sink(void*, record const& rec) noexcept
```
ran-yuan-rui, line 35: this function has the same name within one in logging.cpp

PR #176, opened 2026-07-31
fix: identify GPU by PCI bus id when probing hw decompression

aminaramoon:
## Description

1. move kvikio under cudf targets
2. create 2 phase discovery for runtime attributes to avoid initializing cudaContext.

### DIFF
--- a/src/memory/topology_discovery.cpp
+++ b/src/memory/topology_discovery.cpp
@@ -5,6 +5,8 @@
 
 #include <cucascade/memory/topology_discovery.hpp>
 
+#include <cuda.h>
+
 #include <dlfcn.h>
 #include <ifaddrs.h>
 #include <nvml.h>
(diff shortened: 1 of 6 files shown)


src/memory/topology_discovery.cpp
```diff
@@ -61,25 +58,129 @@ void report_nvml_error(nvmlReturn_t result, std::string const& context)
...
- * @brief Query whether a CUDA device supports hardware-accelerated decompression.
+ * @brief Minimal subset of the CUDA driver API used for device capability queries.
+ *
+ * Resolved with dlopen/dlsym rather than linked, so this library keeps loading on
+ * hosts without an NVIDIA driver (matching how NVML is treated here) and so the
+ * topology-only build needs no CUDA runtime or RMM dependency.
+ *
+ * `CUdevice` and `CUresult` are spelled as `int` to avoid pulling in `<cuda.h>`:
+ * `CUdevice` is a typedef for `int` and `CUresult` is an int-sized enum, so both
+ * match the driver ABI.
```
wence-: This is only necessary if you want to be able to compile without `cuda.h`. Is that required?

src/memory/topology_discovery.cpp
```diff
@@ -61,25 +58,129 @@ void report_nvml_error(nvmlReturn_t result, std::string const& context)
...
+Fn load_symbol(void* handle, char const* name)
 {
-  try {
-    rmm::cuda_set_device_raii set_device{rmm::cuda_device_id{static_cast<int>(cuda_ordinal)}};
-    return rmm::detail::hwdecompress::is_supported();
-  } catch (...) {
+  void* symbol = dlsym(handle, name);
+  Fn fn{};
+  if (symbol != nullptr) { std::memcpy(&fn, &symbol, sizeof(fn)); }
+  return fn;
```
wence-: nit, you should use `dlerror()` to clear any error state before calling `dlsym` and then also inspect `dlerror()` afterwards to see if you got something wrong.

You can also avoid the (probably elided) memcpy with:

```
template<typename Fn>
void load_symbol(Fn& fn, void* handle, char const* name) {
  ::dlerror();
  fn = reinterpret_cast<fn>(dlsym(handle, name);
  char const *err = ::dlerror();
  if (err) { throw(...); }
}
```

src/memory/topology_discovery.cpp
aminaramoon: I guess topology discovery could, but you do have a point as cucascade has cuda as a dependency. so technicallyt it will never happen. I can change that.

.cache/.gitignore
```diff
@@ -0,0 +1 @@
+clangd/
```
wence-: should .cache itself be gitignored?

wence- (approved):

.cache/.gitignore
aminaramoon: oh gosh, you are right! not sure how that slipped in.

mbrobbel:
/ok to test [988048f](https://github.com/NVIDIA/cuCascade/pull/176/commits/988048fe1b2fc420f796412e0d1c0be20c36d437)

src/memory/topology_discovery.cpp
```diff
@@ -995,17 +993,37 @@ bool topology_discovery::discover(NetworkDeviceVerification net_verification)
   for (size_t visible_idx = 0; visible_idx < visible_indices.size(); ++visible_idx) {
     size_t nvml_idx = visible_indices[visible_idx];
     if (nvml_idx >= nvml_gpus.size()) { continue; }
-    auto gpu                       = nvml_gpus[nvml_idx];
-    gpu.id                         = static_cast<unsigned int>(visible_idx);
-    gpu.hw_decompression_available = query_hw_decompression(gpu.id);
+    auto gpu = nvml_gpus[nvml_idx];
+    gpu.id   = static_cast<unsigned int>(visible_idx);
+    // Runtime attributes are populated below only when explicitly requested,
+    // so that plain discovery never initializes a CUDA context.
```
pentschev:
```suggestion
```

pentschev: Seems unnecessary.

pentschev (approved):
Approving based on verification of CUDA initialization, I have been able to confirm that no distortion of `CUDA_VISIBLE_DEVICES` occur due to the early CUDA context initialization, which fixes the issue we had with RapidsMPF's use of topology discovery and allows us to resume using cuCascade's `main` branch during development.

Thanks @aminaramoon .

PR #177, opened 2026-07-31
perf(cudf): skip uploading all-valid null masks in host->GPU reconstruction

felipeblazing:
## What

Two commits:

1. **NVTX phase instrumentation** for `convert_host_fast_to_gpu` — named ranges around each conversion phase (presync, stream acquisition, column reconstruction, device alloc, null-mask sync, offsets cast, batch flush, final sync). Zero-cost without a profiler; with nsys it yields a per-phase wall-time breakdown of every host→GPU conversion.
2. **Skip all-valid null masks** during column reconstruction. Masks upload via a synchronous per-column memcpy + stream sync (cudf factories read them at construction, so they cannot join the async copy batch). When `null_count == 0` the mask is semantically a no-op — skip the upload and construct the column non-nullable.

## Why / measurements

Profiling TPC-H SF1000 host-pinned scans on GB300 (all columns all-valid): the mask syncs accounted for ~10.4 ms of every 39 ms five-GB conversion (~975 blocking round-trips and 5.2 GB of mask bytes per q9 iteration), amplified by pooled-stream aliasing across converting threads. With the fix, per-conversion wall time drops ~2.5x in combination with larger host block sizes, q9 hot improves 3.5%, and copy-engine duty within conversion windows reaches ~89% (link-bound). Query results are byte-identical.

## Semantics caveat for review

Reconstructed all-valid columns become **non-nullable** instead of nullable-with-all-valid-mask. cudf treats these as equivalent for computation, but strict column-equality checks that compare nullability flags could notice. If that is a concern, the alternative is batching mask uploads into the existing `cudaMemcpyBatchAsync` flush for the column constructors that don't read mask contents (fixed-width/LIST/STRUCT) — happy to rework in that direction.

### DIFF
--- a/src/cudf/representation_converter_builtins.cpp
+++ b/src/cudf/representation_converter_builtins.cpp
@@ -971,9 +971,11 @@ static rmm::device_buffer alloc_and_schedule_h2d(
   rmm::device_async_resource_ref mr,
   BatchCopyAccumulator& batch)
 {
-  nvtxRangePushA("hg:dev_alloc");
-  rmm::device_buffer buf(size, stream, mr);
-  nvtxRangePop();
+  rmm::device_buffer buf;
+  {
+    nvtx_scope alloc_range{"hg:dev_alloc"};
+    buf = rmm::device_buffer(size, stream, mr);
+  }
   if (size == 0) { return buf; }
   if (alloc.size() == 0) {
     throw std::invalid_argument(
@@ -1017,9 +1019,11 @@ static rmm::device_buffer alloc_and_copy_h2d_sync(
   rmm::device_async_resource_ref mr)
 {
   nvtx_scope range{"hg:nullmask_sync"};
-  nvtxRangePushA("hg:dev_alloc");
-  rmm::device_buffer buf(size, stream, mr);
-  nvtxRangePop();
+  rmm::device_buffer buf;
+  {
+    nvtx_scope alloc_range{"hg:dev_alloc"};
+    buf = rmm::device_buffer(size, stream, mr);
+  }
   if (size == 0) { return buf; }
 
   const std::size_t block_size = alloc.block_size();
@@ -1197,9 +1201,10 @@ std::unique_ptr<idata_representation> convert_host_fast_to_gpu(
   // host_data_representation is flushed before we read the pinned host blocks.
   // The caller's stream may be bound to a non-target device under multi-GPU;
   // synchronize is safe across devices.
-  nvtxRangePushA("hg:presync");
-  stream.synchronize();
-  nvtxRangePop();
+  {
+    nvtx_scope presync_range{"hg:presync"};
+    stream.synchronize();
+  }
 
   rmm::cuda_set_device_raii device_guard{rmm::cuda_device_id{target_memory_space->get_device_id()}};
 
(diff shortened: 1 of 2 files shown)


felipeblazing:
/ok to test [3fbe5c8](https://github.com/NVIDIA/cuCascade/pull/177/commits/3fbe5c81f9c47f6393d706f4e02c0f546bdc5bca)

src/cudf/representation_converter_builtins.cpp
```diff
@@ -962,7 +971,9 @@ static rmm::device_buffer alloc_and_schedule_h2d(
   rmm::device_async_resource_ref mr,
   BatchCopyAccumulator& batch)
 {
+  nvtxRangePushA("hg:dev_alloc");
   rmm::device_buffer buf(size, stream, mr);
```
aminaramoon: if you get an oom, you will never pop this range.

src/cudf/representation_converter_builtins.cpp
```diff
@@ -1005,7 +1016,10 @@ static rmm::device_buffer alloc_and_copy_h2d_sync(
   rmm::cuda_stream_view stream,
   rmm::device_async_resource_ref mr)
 {
+  nvtx_scope range{"hg:nullmask_sync"};
+  nvtxRangePushA("hg:dev_alloc");
```
aminaramoon: same comment here

src/cudf/representation_converter_builtins.cpp
```diff
@@ -1178,30 +1197,40 @@ std::unique_ptr<idata_representation> convert_host_fast_to_gpu(
...
   stream.synchronize();
+  nvtxRangePop();
 
   rmm::cuda_set_device_raii device_guard{rmm::cuda_device_id{target_memory_space->get_device_id()}};
 
   // Acquire a target-bound stream from the target memory_space's stream pool.
   // Using the caller's stream for the H2D batch under a target-device RAII
   // guard raises cudaErrorInvalidValue when stream and current device belong
   // to different CUDA contexts (multi-GPU case).
+  nvtxRangePushA("hg:acquire_stream");
```
aminaramoon: same

src/cudf/representation_converter_builtins.cpp
```diff
@@ -1178,30 +1197,40 @@ std::unique_ptr<idata_representation> convert_host_fast_to_gpu(
   // host_data_representation is flushed before we read the pinned host blocks.
   // The caller's stream may be bound to a non-target device under multi-GPU;
   // synchronize is safe across devices.
+  nvtxRangePushA("hg:presync");
```
aminaramoon: same

aminaramoon (changes requested):

src/cudf/representation_converter_builtins.cpp
felipeblazing: Fixed in `c95591d` — all seven raw push/pop pairs (not just the four flagged) now use the RAII `nvtx_scope`, so an OOM between push and pop unwinds the range instead of corrupting the thread's NVTX stack.

src/cudf/representation_converter_builtins.cpp
felipeblazing: Fixed in `c95591d` (same RAII conversion).

src/cudf/representation_converter_builtins.cpp
felipeblazing: Fixed in `c95591d` (same RAII conversion).

src/cudf/representation_converter_builtins.cpp
felipeblazing: Fixed in `c95591d` (same RAII conversion).

aminaramoon (approved):

PR #178, opened 2026-08-03
feat(io): add REST performance instrumentation

ran-yuan-rui:
## What

Implements the REST perf-instrumentation follow-up left out of #172. Sirius consumes these counters in its S3 tests and benchmark JSON.
Add loopback HTTP range server for REST conformance tests

## Counter behavior

- Retry, terminal-failure, device-stream-sync, and payload-byte counters are always enabled. Payload bytes count HTTP response-body bytes across attempts, including retries.
- `chunk_get`, `queue_wait`, `h2d_observed`, `blocking_host_get`, and `ttfb_ns` are recorded only when `perf_instrumentation` is enabled. The disabled instrumentation path adds no clock reads.
- Despite its retained name, `ttfb_ns` measures GET submission to completion of the first completed GET, not time to the first byte on the wire.
- Counters use relaxed atomics. Snapshots can be read while reactors are running without taking reactor locks, but exact comparisons should be made after outstanding requests finish.
- `rest_perf_snapshot` is source-pin API rather than a stable binary layout. New fields are appended without reordering existing fields.
- cuCascade exposes only the C++ config field. YAML wiring remains the responsibility of the embedding application.

## Validation

- Standalone: `pixi run -e cuda-12-stable build` / `test` — 4/4 ctest targets, `cucascade_io_tests` 103 cases / 1169 assertions green, `[rest][perf]` 20 cases / 149 assertions.
- Consumer-side: the Sirius integration branch builds against this branch, and its full S3 integration suite and S3 benchmark pass with the perf coverage restored (the counters are consumed by its tests and its benchmark JSON).

### DIFF
--- a/include/cucascade/io/rest/config.hpp
+++ b/include/cucascade/io/rest/config.hpp
@@ -87,6 +87,12 @@ struct config {
   std::chrono::milliseconds retry_jitter{50};
   bool honor_retry_after{true};
 
+  /// When set, the reactor records the per-chunk micro timings (chunk_get,
+  /// queue_wait, ttfb, h2d_observed) into its perf counters.  The retry,
+  /// terminal-failure, device-stream-sync and payload-byte counters are always
+  /// recorded, independent of this flag.
+  bool perf_instrumentation{false};
+
   /// Suffix-range window (bytes) for the parquet footer probe
   /// (@c open_hint::parquet_footer_probe): one `Range: bytes=-N` GET resolves the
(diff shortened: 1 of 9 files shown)


felipeblazing:
/ok to test [3eda942](https://github.com/NVIDIA/cuCascade/pull/178/commits/3eda942a4c5a391b27bc4137241995674b308428)

aminaramoon (approved):

PR #180, opened 2026-08-05
feat(io): capture object ETags and expose a validation tag on io_object 

ran-yuan-rui:
Porting https://github.com/sirius-db/sirius/pull/1401 IO part from Sirius.

## Problem

Embedders that layer a validating cache above `cucascade::io` need a per-open validation tag for the objects they read. DuckDB >= 1.5.5 is a concrete consumer: its external file cache refuses to cache any file whose filesystem reports no validation metadata, so an embedder running on `cucascade::io` silently loses all caching for object-store reads — every read goes back to the network (sirius-db/sirius#1401 is the downstream fix this generalizes).

## Change

Capture the HTTP ETag on the two blocking metadata requests the open path already makes, and expose it through the `io_object` seam:

- `io_object` gains `validation_tag()` — a defaulted virtual returning an empty view, documented as an opaque, equality-only cache validator (quotes and any weak `W/` prefix preserved; explicitly NOT a precondition token for `If-Match`/`If-Range`). Backends that don't override it are unaffected: an empty tag means "not validatable", which callers must treat as cache-degraded, never wrong-bytes.
- The suffix-range footer probe records the ETag of the verified 206 only; a 200, 416, or malformed 206 discards it (those responses also discard the probe itself). The tag resets on every HTTP status line, so interim responses within one transfer leave no residue, and a failed retry attempt's tag dies with its attempt.
- New `rest_reactor::head_object()` returns `head_object_result{object_size, etag}`; `head_object_size()` remains as a size-only wrapper with an unchanged signature. The HEAD path uses its own header callback (`head_capture`) so the async data-GET path parses nothing it does not consume.
- `rest_ioctx` threads the tag into `rest_io_object` on the HEAD open, probe open, and probe-fallback paths. The `known_size` open (size from ListObjectsV2) stays tag-less and network-free, and that is pinned by a test.

No additional network I/O: capture rides existing requests, and the async column-chunk read path is untouched. Per-open overhead is one extra header match per relevant header plus one `std::string` per REST object.

## ABI disclosure

This PR changes the `io_object` vtable, the public layouts of `footer_probe` and `rest_io_object`, and the `rest_io_object` constructor symbols. Consumers must rebuild against the same source pin. The existing `head_object_size` symbol and calling convention remain unchanged, but the PR as a whole is not ABI-compatible with previously built binaries.

## Tests

`test/io/rest/test_rest_validation_tag.cpp` (+8 cases, 103 -> 111 in `cucascade_io_tests`) on the loopback range server, which now optionally emits ETags, per-fault-response tags, and interim 1xx header blocks: quoted/weak/empty-entity tags preserved verbatim; missing tag stays empty end-to-end; retry attempts isolated in both directions (fresh tag wins; a failed attempt's tag never leaks); 200/416/malformed-206 discard then HEAD-fallback supplies the tag; interim-response tags do not leak into the final response; every network open path threads the tag; capture adds zero requests; known-size opens stay tag-less with zero requests. All pre-existing cases pass unchanged.

### DIFF
--- a/include/cucascade/io/rest/rest_reactor.hpp
+++ b/include/cucascade/io/rest/rest_reactor.hpp
@@ -122,9 +122,8 @@ struct head_object_result {
 /**
  * @brief Concrete @c io_object backed by a RESTful object-store key.
  *
- * Passive bag of identity: the original URL/path (also the cache id), the
- * bucket + key the reactor authorizes against, and the object size discovered
- * by a one-time HEAD at construction.  Does no I/O of its own.
+ * Stores the object identity and metadata captured when it was opened.
+ * Does no I/O of its own.
  */
 class rest_io_object : public io_object {
  public:
--- a/include/cucascade/io/rest/rest_reactor.hpp
+++ b/include/cucascade/io/rest/rest_reactor.hpp
@@ -104,6 +104,15 @@ struct footer_probe {
   std::size_t object_size{0};
   std::size_t window_lo{0};
   shared_byte_span bytes;
+  /// ETag from the verified 206, quotes preserved; empty otherwise.
+  std::string etag;
+};
+
+/// Result of a blocking HEAD: the object's size plus its ETag when the server
+/// sent one (quotes preserved, empty otherwise).
+struct head_object_result {
+  std::size_t object_size{0};
+  std::string etag;
 };
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


felipeblazing:
/ok to test [ae0aeb2](https://github.com/NVIDIA/cuCascade/pull/180/commits/ae0aeb262683135e6d92ea10f51fcb53fac3c3a9)

aminaramoon (approved):

ran-yuan-rui:
@felipeblazing and @aminaramoon the build error already fix at PR 179.
but as I will need help to rerun the test - please help here as I do not have permission

PR #181, opened 2026-08-06
Feature - Declare the S3RDMA ioctx and add backend integration seams

ran-yuan-rui:
## Summary

This PR declares the cuCascade interfaces needed by the S3RDMA backend. It does not implement the S3RDMA transport.

The transport implementation remains in Sirius. The non-constructible `s3rdma_ioctx` header links to the [detailed design document](https://github.com/sirius-db/sirius/blob/dev/experimental/s3-rdma-transport-design.md).

## Changes

- Add `io_context_type::s3rdma`.
- Add a non-constructible `s3rdma_ioctx` declaration.
- Add `rest::object_store_lister` and compose it into `rest_ioctx` behind the existing listing methods.
- Add `io_context_registry::replace_ioctx()`.
- Add `templated_ioctx::on_device_dispatch_failure()`.

### `replace_ioctx()`

cuCascade initially assigns `s3://` paths to the REST backend. When S3RDMA is selected, Sirius uses this API before the first path lookup to replace the REST registration with the S3RDMA checker and factory.

The replacement runs under the registry lock and leaves one explicit claimant for `s3://`. It changes registration metadata only; it does not create, stop, or migrate an I/O context.

### `on_device_dispatch_failure()`

`templated_ioctx` normally converts synchronous device-dispatch exceptions into errored futures. S3RDMA needs to check whether such a failure came from a sticky CUDA context error before returning it as a normal request failure.

The new protected hook provides that check point. Its default implementation is a no-op, so existing backends keep their current behavior.

## Testing

- Added registry replacement tests.
- Added dispatch-failure hook tests.
- Added listing capability tests.
- Full cuCascade test suite passes (`cucascade_io_tests` 113 → 134 cases).


ran-yuan-rui:
@aminaramoon @felipeblazing Please help to review and start to test

include/cucascade/io/object_store_listing.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include <optional>
+#include <string_view>
+
+namespace cucascade::io {
+
+/// Listing capability of an object-store backend.  A glob / LIST layer
+/// depends on this interface, not a concrete ioctx type.  Listing is prefix
+/// resolution on the control plane; it is independent of how the data plane
+/// reads the resolved keys.
+class object_store_listing {
```
aminaramoon: reactors should definitely not extend this. this shold be a resource inside the reactor context, right ?

aminaramoon (changes requested):

include/cucascade/io/object_store_listing.hpp
ran-yuan-rui: Agreed on composition. Fixed in b1fc466. The interface has been removed, and the shared ListObjectsV2 logic now lives in rest::object_store_lister. rest_ioctx owns the lister and supplies rest_reactor::list_page; S3RDMA will supply its control-plane LIST call.
The lister stays on the ioctx because it operates over the reactor pool. reactor_context remains limited to resources shared by individual reactors.

aminaramoon (approved):

mbrobbel:
/ok to test [007cd44](https://github.com/NVIDIA/cuCascade/pull/181/commits/007cd4464f5736958c90585a52776613639cffc3)

PR #182, opened 2026-08-12
fix(data): order GPU representation reads after writers

kevkrist:
## Summary

This change orders GPU-table reads performed by `clone()` and `release_table()` after the
representation's recorded writer work, without host-synchronizing the consumer stream.

- Owned and view-backed GPU representations record a writer event during construction, including
  when the writer is the CUDA default stream.
- `clone()` waits for the source writer event on the clone stream before enqueueing its deep copy.
  The cloned representation records its own writer event and `clone()` may return while that stream
  is still pending.
- `release_table()` waits for the source writer event on the release stream before an owned table is
  handed off or a view-backed table is materialized. It may likewise return while the release stream
  is still pending.
- The API documentation now states the required device, stream, and source-lifetime preconditions.
- Regression tests cover default-stream event creation and gated producer ordering for both
  `clone()` and `release_table()`.

## Failure example: `clone()`

Assume a source allocation initially contains zeroes:

1. Producer stream **P** is paused, with a write of `0x5a` and source writer event **Wsrc** queued
   behind the pause.
2. Clone stream **C** calls `clone()` while **P** is still paused.
3. Before this fix, **C** could enqueue its copy without waiting for **Wsrc**, so the copy was free
   to read the old zeroes instead of `0x5a`.
4. With this fix, **C** first waits for **Wsrc**, then enqueues the copy and records the clone's own
   writer event **Wclone**.

The resulting dependency chain is:

```text
P: source write -> Wsrc
                    |
                    v
C:              source-to-clone copy -> Wclone
```

`clone()` remains asynchronous: it can return while **C** is pending. The caller must keep the
source alive and unmodified until the clone-stream work completes. A consumer of the clone must
wait for **Wclone** before reading it from another stream.

## Failure example: `release_table()`

Before this fix, `release_table()` could hand an owned table to another stream, or begin
materializing a view-backed table there, without ordering that stream after the producer. The
release stream could therefore observe stale or partially written source data.

The fixed path inserts a wait for the representation's writer event on the release stream before
the table is handed off or materialized. The operation remains asynchronous and does not
synchronize the release stream. For a view-backed representation, the external owner must continue
to keep the viewed allocations alive and unmodified until the queued materialization completes.

## Scope

This patch does not register clone work as an asynchronous `data_batch` reader, add a
`wait_until_ready()` API, change the mutable clone/`clone_to()` APIs, synchronize clone or release
streams, or add exception-path stream draining. Those lifetimes remain governed by the documented
caller preconditions.

### DIFF
--- a/include/cucascade/cudf/gpu_data_representation.hpp
+++ b/include/cucascade/cudf/gpu_data_representation.hpp
@@ -118,13 +118,15 @@ class gpu_table_representation : public idata_representation {
   std::size_t get_uncompressed_data_size_in_bytes() const override;
 
   /**
-   * @brief Create a deep copy of this GPU table representation.
+   * @brief Create an independently owned copy of this GPU table
    *
-   * The cloned representation will have its own copy of the underlying cuDF table,
-   * residing in the same memory space as the original.
+   * Orders the copy after the recorded writer event, or synchronizes the source device if no event
+   * is available. The copy uses this memory space's default allocator on @p stream, and the method
+   * synchronizes @p stream before returning. A stream with a non-null handle is recorded as the
+   * result's writer stream.
    *
-   * @param stream CUDA stream for memory operations
-   * @return std::unique_ptr<idata_representation> A new gpu_table_representation with copied data
+   * @param stream Stream on this representation's device used for the copy
+   * @return Independently owned copy in the same memory space
    */
   std::unique_ptr<idata_representation> clone(rmm::cuda_stream_view stream) override;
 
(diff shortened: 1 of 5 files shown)


mbrobbel:
/ok to test [e62614a](https://github.com/NVIDIA/cuCascade/pull/182/commits/e62614a216f14f9039e73c76c37214cfcd389730)

mbrobbel:
/ok to test [24496d8](https://github.com/NVIDIA/cuCascade/pull/182/commits/24496d8a149e4daba0d5fd566114a4518c12c4c1)

PR #184, opened 2026-08-14
feat(data): wait for asynchronous batch readers

9prady9:
## Summary

- add `read_only_data_batch::record_reader_event(stream)` so consumers can publish asynchronous GPU reads before releasing their shared lock
- retain all outstanding reader completions in reusable, per-device CUDA event pools
- make blocking mutable acquisition wait for every recorded reader while keeping `try_to_mutable()` non-blocking
- preserve the requested no-op behavior for host and disk representations, with a destructor safety fallback for the final batch owner

## Why

A reader can enqueue an asynchronous copy from a non-owning GPU view and release its read lock before the copy finishes. The batch can then be reclaimed and its memory reused, silently corrupting the in-flight copy. Writer events only order readers after writers; they do not order reclaim after readers.

This change records the inverse dependency on the batch. Any path that obtains mutable access—including replacement, conversion, rebind, and raw mutable access—crosses the same reader-event barrier before it can reuse the representation.

## Correctness details

- concurrent readers retain independent events; a later fast reader cannot overwrite an earlier slow dependency
- event pools are keyed by the CUDA device that owns each reader stream
- failed event registration synchronizes the already-enqueued reader stream before propagating, preventing an untracked read from escaping the lock lifetime
- `readonly_to_mutable()` uses the same barrier, while `try_to_mutable()` queries events and returns `nullopt` when any remain pending

## Test plan

- [x] `pixi run pre-commit run --files include/cucascade/data/data_batch.hpp src/data/data_batch.cpp test/data/test_data_batch.cpp`
- [x] warnings-as-errors build of `cucascade_tests`
- [x] warnings-as-errors build of `cucascade_cudf_tests`
- [x] regression covers two concurrent readers, both mutable transition paths, poison-on-reclaim ordering, completed-event reuse, and non-GPU no-op behavior
- [x] hidden multi-GPU regression covers event-pool reuse after representation replacement on another device
- [ ] run `cucascade_cudf_tests '[reader_event]'` on a GPU host; this development host has no usable NVIDIA driver and returns `cudaErrorNoDevice`


bwyogatama:
Heads-up: pushed 3 commits onto this branch (maintainer edit) rather than continuing our overlapping PR #186 — after comparing the two, this PR's design is the better base (per-device pools, single barrier at mutable acquisition), so we consolidated here. What was added:

1. **`8cc3646` — `release_table(stream)` now rebinds the owned table's buffers to the passed stream** before transferring ownership. This is the ownership-transfer half of the reclaim-safety story that reader events can't cover (stolen buffers leave the batch, so there's nothing to record on): pre-fix, released tables stayed dealloc-bound to the internal conversion-pool stream, and freeing a column mid-read let `cudaMallocAsync` recycle the block under in-flight kernels — the root cause of the Sirius TPC-H Q18 SF1000 corruption (confirmed via compute-sanitizer: 258 stream-ordered UAF hazards). The `@pre` documents the drained-previous-stream precondition and points to `record_reader_event` + the acquisition barrier as the sanctioned remedy for in-flight foreign reads.
2. **`c87534a` — tests**: a `release_table` suite (first coverage of that function: recursive stream-binding assertions incl. nested columns with an anti-vacuity control, a 20-iteration allocator-reuse UAF stress, view-branch semantics) plus 4 lean additions to the reader-event suite covering gaps: concurrent shared-lock recording (8 threads), the swap-compaction branch of `recycle_completed_reader_events` (completed tail behind a gated pending head — the fully-drained recycle test never reaches it), default-stream recording, and the destructor fallback (previously untested; measured to genuinely block ~100 ms on a gated reader).
3. **`1b18e4b` — required build fix**: the `#include <cucascade/cuda/event.hpp>` in `data_batch.hpp` makes `cucascade::cuda` visible inside `namespace cucascade`, hijacking unqualified `cuda::stream_ref` lookup in `reservation_aware_resource_adaptor_impl.hpp:203,213` (and latently `small_pinned_host_memory_resource.hpp:105,114`) — the Sirius build fails in 5+ TUs without qualification to `::cuda::stream_ref`. Applied the 4-line fix; any embedder with those include orders would hit this.

**First hardware validation** (your checklist's open box — this branch on a GB300): all `[reader_event]` cases pass (154 assertions incl. the additions); the acquisition barrier measurably host-blocks (~1.02 s per section = the 1 s watchdog release, i.e. not vacuous); full `cucascade_cudf_tests` suite green (190 cases). Caveat: the `[.multi-device]` test self-skips on a single-GPU box, so the per-device pool claim is still hardware-unverified.

A few observations from running/reading, for your consideration (none blocking):
- `record_reader_event`'s non-GPU no-op leaves pinned-HOST reads unprotected — an in-flight H2D DMA from pinned memory is still a UAF if the host buffer is reclaimed; today's callers only record on GPU-tier batches, but a host-tier recorder gets silent non-protection rather than an error.
- The barrier holds `_reader_events_mutex` across a host `cudaEventSynchronize`; benign under current lock discipline (recorders hold the shared lock), worth a comment so future lock-free recorders don't serialize against a blocked reclaimer.
- `try_to_mutable() == nullopt` now also means "readers pending", not just lock contention — fine for skip-and-retry callers (Sirius's prefetcher), but it's a contract change.
- In the recycle loop, `query() == error` counts as pending — so a sticky CUDA error makes `try_to_mutable` decline forever while `to_mutable` throws; reasonable, but the asymmetry deserves a comment.

Downstream: Sirius PR sirius-db/sirius#1566 now adopts `record_reader_event` at its filtered scan serving sites (first production recorders) and pins its submodule to this branch. Our PR #186 will be closed in favor of this one.

bwyogatama (approved):

9prady9:
Thank you for pushing this ahead over the weekend :)

@bwyogatama I found couple of concerns, looking into those. Hence didn't merge yet.

9prady9:
@bwyogatama — two items from the review turn on design intent rather than a clear defect, so I'd like your call before changing anything. The rest of the findings I'm handling separately.

### 1. `to_mutable()` blocks on CUDA events while holding the exclusive lock

`src/data/data_batch.cpp:222-228` takes the exclusive `_rw_mutex`, then blocks in `synchronize_reader_events()`. `readonly_to_mutable()` (:249-259) has the same ordering.

The trigger originally reported was a CUDA host callback that reenters `to_read_only()`. I'd discount that one: `cudaLaunchHostFunc` explicitly forbids host functions that "perform any synchronization that may depend on outstanding CUDA work not mandated to run earlier," and nothing in-tree does it.

But there's a variant that needs no callback and isn't prohibited by anything. If a reader stream `S` has an enqueued `cudaStreamWaitEvent(S, E)` ahead of the read, and `E` is recorded by a thread that must first take a read lock on the same batch, then: the mutator parks in `synchronize_reader_events()` holding the exclusive lock → the recording thread blocks in `to_read_only()` → `S` never drains → the mutator waits forever.

**Question: keep the wait under the lock?**

Fixing it means waiting outside `_rw_mutex` and re-checking `reader_events_complete()` under it in a retry loop. The "no mutable accessor exists while a reader event is pending" invariant survives, because every mutator re-checks under the exclusive lock immediately before constructing the accessor. The cost is that acquisition becomes reader-preferring — continuous reader arrival can starve it. In practice glibc's `shared_mutex` blocks new readers behind a queued writer so it should converge, but it's no longer bounded.

The cheaper alternative is to make the current behavior contractual: document on `record_reader_event()` that work registered as a reader must not depend on any thread acquiring a lock on the same batch. Honest, but it pushes the constraint onto callers.

### 2. `~data_batch()` synchronizes reader events

`src/data/data_batch.cpp:49-55`. The comment there calls this "load-bearing when the final read-only accessor owns the last `shared_ptr`," and the test at `test/data/test_data_batch.cpp:1513` asserts the blocking behavior — so this reads as deliberate. It was flagged as conflicting with the original requirement that releasing a read lock stays nonblocking.

**Question: was that requirement intentionally relaxed for the last-reference path?**

If yes, I'll add a note to the accessor docs so callers know that dropping the final reference can block. If not, the alternative is async retirement — keeping the representation alive off-thread until the registered events retire — which is a larger change I'd rather scope separately.

bwyogatama:
@9prady9 Thanks for continuing this amazing PR!

For the first one, I don't think there is a case where a reader of a batch is waiting for an event from another reader, so i think we are safe from that deadlock. However, I think we can still wait for the event before acquiring mutable lock just to make it less prone for things like this. I think the main concern if we are doing this is starvation: we can't downgrade cause new readers keep coming. But, we probably shouldn't downgrade anyway if there are still new readers

For the second one, I think we can just make an exception and block on the final reference

Those are my take, let me know if you have concerns or questions!

bwyogatama:
Both points resolved in the latest push (`d18d96d`, rebased onto your `b5423e9`):

**1. Wait moved before lock acquisition** — implemented your proposed fix: both `to_mutable()` and `readonly_to_mutable()` now drain reader events with no lock held, then acquire the exclusive lock and re-check `reader_events_complete()` under it, retrying if a new reader recorded in the interim. The no-mutable-access-while-readers-pending invariant is unchanged (every path re-checks under the lock before constructing the accessor); `try_to_mutable()` needed no change. Header docs updated.

Two notes on the decision. First, on constructibility: with plain cudaEvent semantics the no-callback variant can't actually park a stream — `cudaStreamWaitEvent` on a never-recorded event is a functional no-op, and in-tree readers only wait on writer events recorded pre-publication — so the cycle additionally requires `cuStreamWaitValue*`, external semaphores, or a prohibited host callback, none of which appear here. We moved the wait anyway: it future-proofs against those primitives arriving, and the reader-preference it introduces is arguably the right policy — the main mutable-acquisition path is downgrade/eviction, and a batch with readers in flight is a poor eviction candidate anyway. On starvation: unbounded in theory, but converging in practice since new readers queue behind a parked writer on glibc.

Composes well with your `32231fd`: a failed event query now surfaces out of the re-check instead of wedging the loop.

**2. Destructor sync: yes, intentionally relaxed for the last-reference path.** The final-owner drop is the one reclaim with no later reclaimer to pay the wait, so the destructor pays it — and this is already documented on the accessor (`record_reader_event`'s `@note`: "the shared-lock release itself never synchronizes; if this accessor owns the final shared_ptr to the batch, batch destruction synchronizes as a lifetime-safety fallback") and pinned by the destructor test. No code change; agreed async retirement is the right shape if this ever needs to be non-blocking, scoped separately.

Validation on GB300 at `d18d96d`: full `cucascade_cudf_tests` 2,373 assertions pass (incl. your new cross-device rejection tests and the barrier-blocking test, which still measures the host wait from its new pre-lock position), and the downstream Sirius integration tags pass (sirius-db/sirius#1566 bumped to this tip).

PR #186, opened 2026-08-15
fix(cudf): stream-ordered reclaim safety — release_table stream rebind + consumer-event API

bwyogatama:
## Problem

Buffers of GPU tables produced by the built-in host→GPU converter are allocated on an internal `memory_space` pool stream. Their stream-ordered frees (`cudaFreeAsync`) therefore execute on an idle, foreign, shared stream — unordered against reads still in flight on the consumer's stream. When a consumer drops such a buffer mid-read (e.g. a `cudf::cast` column replacement), the async allocator recycles the block into a concurrent conversion's H2D write while the reader kernel is still consuming it.

Downstream (Sirius), this caused silent, transient data corruption: TPC-H SF1000 Q18 returned wrong aggregates in ~39% of warm pinned-host executions; `compute-sanitizer memcheck` captured 258 identical stream-ordered use-after-free hazards at the consuming cast. ASan/TSan are structurally blind to this class (GPU-timeline race; pool memory is never poisoned).

## Changes

1. **`gpu_table_representation::release_table(stream)` now honors its stream parameter** — the owned-table branch rebinds every column's buffers (recursively, incl. null masks and children) to the passed stream before transferring ownership, so the eventual frees are ordered behind the new owner's reads. The view branch already deep-copies on the passed stream. The drained-previous-stream precondition is documented (`@pre`), and holds for all built-in converters (they synchronize before publishing).
2. **Consumer-event API on `data_batch`** — the mirror of the existing writer event: readers call `record_consumer_event(stream)` after enqueuing reads; reclaimers call `await_consumers(stream)` (device-side waits, no host stall) or observe `consumers_done()`. Backed by a new pooled-event utility (`cuda::event_pool`, `cudaEventDisableTiming`, thread-safe, recycling). Wired into the two in-library reclaim sites: `install_converted_representation` (including a host-sync fallback for non-GPU-tier swaps, where the stream-sync tail doesn't run) and `set_data`.
3. **Namespace hygiene** — qualified 4 pre-existing unqualified `cuda::stream_ref` uses that break once `cucascade::cuda` enters those TUs' lookup.

## Tests

- `test/data/test_consumer_events.cpp` (14 cases): API contracts, device-side gating (verified via device flags + `cudaErrorNotReady` probes), 1000-cycle recycling, 8-thread concurrency at both the pool and batch layers, and deterministic poison-on-reclaim regression tests for all three reclaim hooks.
- `test/cudf/test_release_table_stream.cpp` (5 cases): first coverage of `release_table` — recursive stream-binding assertions with an anti-vacuity control block, the motivating 20-iteration allocator-reuse UAF stress, view-branch semantics, and composition with caller-side rebinds.
- Full `cucascade_cudf_tests` suite green on GB300 (203 assertions in the new suites; no regressions elsewhere).

## Downstream validation

Together with sirius-db/sirius#1566, TPC-H SF1000 full-suite (22 queries × 4 iterations × 4 sessions, full untruncated result validation): **352/352 clean**, vs ~77% corrupt sessions before. No performance regression (same-machine controls).

## Note

The core `cucascade_tests` target currently fails to compile on `main` due to a pre-existing Catch2 v2 include (`test/memory/test_reservation_manager_configurator.cpp:29`, introduced in #173) — unrelated to this PR and not fixed here; the new tests live in `cucascade_cudf_tests`.

### DIFF
--- a/include/cucascade/cuda/event_pool.hpp
+++ b/include/cucascade/cuda/event_pool.hpp
@@ -1,0 +1,108 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cucascade/cuda/event.hpp>
+
+#include <rmm/cuda_stream_view.hpp>
+
+#include <mutex>
+#include <vector>
+
+namespace cucascade {
+namespace cuda {
+
+/**
+ * @brief Thread-safe pool of recorded CUDA events tracking outstanding asynchronous work.
+ *
+ * Callers record() an event on the stream where they just enqueued work; later, another
+ * party either enqueues device-side waits on all outstanding events via enqueue_waits()
+ * (no host sync) or blocks the host until they complete via synchronize().
+ *
+ * Events are created lazily with cudaEventDisableTiming and recycled once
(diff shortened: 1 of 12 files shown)


bwyogatama:
Consolidating into #184, which lands the same reader/consumer-event primitive with a better transport (per-device pools, single barrier at mutable acquisition). Everything unique to this PR — the release_table stream rebind, the namespace qualification fix, the release_table test suite, lean reader-event test additions, and first GPU validation — has been pushed onto #184 (commits 8cc3646, c87534a, 1b18e4b; see https://github.com/NVIDIA/cuCascade/pull/184#issuecomment-5312448919). Sirius PR sirius-db/sirius#1566 now tracks #184's branch.

PR #188, opened 2026-08-17
feat(io): batched footer resolve on rest_ioctx

ran-yuan-rui:
## Description

The existing footer-probe path starts with a blocking suffix GET on a short-lived easy handle. The process-wide CURLSH shares DNS and TLS-session state, but not its connection cache, so repeated probes each create a new TCP/TLS connection.

This adds `rest_ioctx::resolve_footer_objects(paths, on_result, stop)`. A per-call curl multi keeps at most `footer_resolve_max_inflight` transfers active and reuses connections while the call is running. Results are delivered one input at a time, serially on the caller's thread, without waiting for the whole batch. This PR adds the cuCascade mechanism only; Sirius scan-preparation wiring will follow separately once the sirius-side io parity settles.

Each entry follows `open_io_object(path, parquet_footer_probe)` behavior. A verified 206 returns the suffix window. Responses that cannot establish a valid suffix window fall back to HEAD. A zero-byte probe window uses HEAD directly. Retries re-authorize each attempt and preserve ETag and perf-counter attribution.

Each input occurrence receives exactly one callback, including duplicate paths. Per-entry failures do not cancel siblings. Cancellation reports `operation_canceled` for entries not yet delivered. If a callback throws, remaining entries are canceled before the first exception is rethrown. Invalid submissions fail before any callback or network request.

The returned `io_object` carries the path, object size, and validation tag, but not the footer bytes. Footer bytes are returned in a separate payload backed by an ioctx-wide `footer_resolve_stash_budget` lease. The reservation is released when the payload is freed. Concurrent calls are FIFO-serialized per ioctx, and `footer_resolve_max_inflight = 0` disables the API.

Supporting changes:

- `exec::admission_control` gains `try_acquire()`, `reserved()`, and `peak_reserved()`.
- `rest_perf_snapshot` reports current and peak footer-payload reservations.
- The loopback server gains keep-alive, per-key counters and response scripts, and a deterministic GET barrier.

The PR also moves `mock_authorizer.hpp` from `include/cucascade/io/rest/` to `test/io/rest/`. It is a test-only fixture that was being installed and included in public API documentation. The move preserves its contents and updates its three in-tree includes; the move itself requires no CMake change. Fresh installs no longer contain the header.

## Validation

The 20 footer-resolve cases, all failing before the mechanism landed, cover:

- suffix/HEAD parity, retries, ETags, and perf counters;
- streamed completion, duplicate inputs, cancellation, and callback failures;
- per-entry error isolation;
- memory-budget accounting, FIFO admission, and payload lifetime;
- connection reuse under a bounded in-flight limit.

`cucascade_io_tests`: 135/135 green, 1624 assertions.

Sirius rebuilt successfully with cuCascade `c0a2607` on `integrate_cucs_io` at `8e7ce16b`. `make s3-test` passed 90 of 92 cases; the two remaining failures are the existing retry-log observability cases tracked in #175. The later test-only header move does not affect that result.


aminaramoon (approved):

ran-yuan-rui:
Merged main to pick up #190 No changes to the PR's own commits. 
Could I get another /ok to test? Thanks for the labels and the first run. @aminaramoon

mbrobbel:
/ok to test [0cba7d8](https://github.com/NVIDIA/cuCascade/pull/188/commits/0cba7d8eeb461ee44c4048d27b7c38fbfc9cdf6f)

PR #189, opened 2026-08-18
fix: pool pinned-slab CUDA events per device

9prady9:
## Summary

Fix cross-device CUDA event reuse in `small_pinned_host_memory_resource`. The resource previously kept one global event pool, so an event created on one GPU could be recorded on another GPU's stream. CUDA rejects that combination; in release builds, the slab could then be returned to the free list without correct stream ordering and reused during an in-flight transfer.

## Changes

- Pool events by the CUDA device that owns the stream.
- Create events on that device and preserve their device ownership with each free slab.
- Fail closed when event creation, recording, waiting, or synchronization fails.
- Synchronize outstanding events before releasing backing storage during teardown.

## Validation

- Formatting, pre-commit checks, the CUDA 13.2 Release build, and tests pass.
- Compute Sanitizer errors from this resource dropped from 2,361 `cudaErrorInvalidResourceHandle` failures to none.
- Multi-GPU behavior was verified on GB200 with CUDA 13.2.

## Context

This was found while debugging intermittent TPC-H query 6/9 correctness failures in Sirius.

### DIFF
--- a/src/memory/small_pinned_host_memory_resource.cpp
+++ b/src/memory/small_pinned_host_memory_resource.cpp
@@ -37,39 +37,107 @@ small_pinned_host_memory_resource::~small_pinned_host_memory_resource()
   // free_lists_ entries are raw pointers into those blocks; no individual cleanup needed.
   // Destroy every CUDA event we own — both idle (pooled) and still attached to a
   // free slab that was never re-allocated.
-  for (auto& event : event_pool_) {
-    if (event != nullptr) { CUCASCADE_ASSERT_CUDA_SUCCESS(::cudaEventDestroy(event)); }
+  bool retain_backing_allocations = has_quarantined_slab_;
+  int prev_device                 = 0;
+  bool const have_prev            = (::cudaGetDevice(&prev_device) == cudaSuccess);
+  auto destroy_on                 = [&retain_backing_allocations](cudaEvent_t event, int device) {
+    if (event == nullptr) { return; }
+    // cudaEventDestroy is device-agnostic, but making the owning device current
+    // keeps the call on the context the event was created in.
+    if (device >= 0) { (void)::cudaSetDevice(device); }
+    // Destroying an incomplete event does not wait for its captured work. Wait
+    // explicitly before owned_allocations_ returns the backing slabs upstream.
+    if (::cudaEventSynchronize(event) != cudaSuccess) {
+      // Fail closed: cudaEventDestroy is non-blocking for incomplete events.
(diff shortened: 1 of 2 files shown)


Jedi18:
/ok to test https://github.com/NVIDIA/cuCascade/pull/189/commits/976247968d60dc9d26125a6551a00ea1c70fef03

9prady9:
/ok to test https://github.com/NVIDIA/cuCascade/pull/189/changes/4fec144a8d1791bdfcde98dfa11e7f4ccf0bf861

Jedi18:
/ok to test https://github.com/NVIDIA/cuCascade/pull/189/commits/4fec144a8d1791bdfcde98dfa11e7f4ccf0bf861

src/memory/small_pinned_host_memory_resource.cpp
```diff
@@ -37,39 +37,107 @@ small_pinned_host_memory_resource::~small_pinned_host_memory_resource()
...
+int small_pinned_host_memory_resource::device_of_stream(::cuda::stream_ref stream) noexcept
+{
+  int device = -1;
+  // A CUDA event is bound to the device current at creation time, and
+  // cudaEventRecord requires the event and stream to be on the same device.
+  // Ask the driver which device this stream belongs to rather than assuming it
+  // matches the calling thread's current device.
+  if (::cudaStreamGetDevice(stream.get(), &device) == cudaSuccess) { return device; }
+  (void)::cudaGetLastError();
+  if (::cudaGetDevice(&device) == cudaSuccess) { return device; }
```
felipeblazing: when would this be invoked? when should a stream not have a device?

src/memory/small_pinned_host_memory_resource.cpp
```diff
@@ -37,39 +37,107 @@ small_pinned_host_memory_resource::~small_pinned_host_memory_resource()
...
+      return nullptr;
+    }
+  }
+
   cudaEvent_t event = nullptr;
   // Timing is not needed; disabling it makes record/wait cheaper.
-  if (::cudaEventCreateWithFlags(&event, cudaEventDisableTiming) != cudaSuccess) {
-    // Best effort: without an event this deallocation loses stream ordering, but
-    // allocation must never throw here. Clear the sticky error and carry on.
+  bool const created = (::cudaEventCreateWithFlags(&event, cudaEventDisableTiming) == cudaSuccess);
```
felipeblazing: how often is this path used? are the pools too small?

bwyogatama (approved):

PR #191, opened 2026-08-18
Add regression coverage for CUDA-context-independent topology discovery

pentschev:
PR #176 keeps default topology discovery independent of CUDA runtime and context state. This PR closes the remaining coverage gap by adding a multi-GPU regression test that verifies default topology discovery does not initialize CUDA or capture temporarily widened `CUDA_VISIBLE_DEVICES`, as well as documenting that default discovery is CUDA-context independent, while runtime-attribute discovery is explicitly opt-in.

### DIFF
--- a/src/memory/topology_discovery.cpp
+++ b/src/memory/topology_discovery.cpp
@@ -5,9 +5,6 @@
 
 #include <cucascade/memory/topology_discovery.hpp>
 
-#include <rmm/cuda_device.hpp>
-#include <rmm/detail/runtime_capabilities.hpp>
-
 #include <dlfcn.h>
 #include <ifaddrs.h>
 #include <nvml.h>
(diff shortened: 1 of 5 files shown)


dhruv9vats (approved):

aminaramoon (approved):

src/memory/topology_discovery.cpp
```diff
@@ -63,23 +60,18 @@ void report_nvml_error(nvmlReturn_t result, std::string const& context)
...
+bool query_hw_decompression()
 {
-  try {
-    rmm::cuda_set_device_raii set_device{rmm::cuda_device_id{static_cast<int>(cuda_ordinal)}};
-    return rmm::detail::hwdecompress::is_supported();
-  } catch (...) {
-    return false;
-  }
+  constexpr int min_hw_decompression_cuda_version = 12080;
+  int driver_version                              = 0;
```
felipeblazing, line 72: not all hardware has a DE engine even if the driver supports it

felipeblazing (changes requested):

src/memory/topology_discovery.cpp
pentschev: Yes, you're right. I realized that and have been discussing this offline as well, I'll link to the existing discussion, in the meantime we should hold merging this.

felipeblazing (approved):

PR #192, opened 2026-08-19
feat!: make `read_only_data_batch` truly const

dhruv9vats:
Make `read_only_data_batch` truly const, disallowing any mutating operations, directly or indirectly.

### DIFF
--- a/include/cucascade/data/data_batch.hpp
+++ b/include/cucascade/data/data_batch.hpp
@@ -31,7 +31,6 @@
 #include <mutex>
 #include <optional>
 #include <shared_mutex>
-#include <stdexcept>
 #include <unordered_map>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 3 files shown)


dhruv9vats:
The [failing test](https://github.com/NVIDIA/cuCascade/actions/runs/35193675750/job/105113072089?pr=192) is a flaky one, because:
- The main thread acquires ro_initial, but t1 releases it; unlocking a mutex from another thread is undefined behavior.
- The 50 ms sleep does not guarantee that t2 attempted or blocked on mutable acquisition.
- The last reader can unlock before t1 logs its release, allowing t2 to log acquisition first and fail the ordering assertion. (This is the cause of the seen failure)
- After thread2_mutable_released is set, t1 can acquire new readers before t2 asserts idle.

So disabling the test for now.

9prady9 (commented):
whats the contract we are trying to adhere here ? for the `readonly_to_mutable` and `mutable_to_readonly` . Later breaks the API, former make it strict read-only. They are not exactly illegal is my question about.

the `const` ness change looks good.

dhruv9vats:
These changes basically lay the groundwork for https://github.com/sirius-db/sirius/issues/1063 but incrementally molding the API. The removal of the conversion API basically mandates that the user go through the idle state before acquiring the alternate state. Going from idle to mutable was a grey area anyways because that meant that anyone with a read only handle could get to and mutate the internal state, which broke the invariant of immutability.

include/cucascade/data/data_batch.hpp
kevkrist: It is a bit strange but I believe these now need to be `const`. E.g., if a read only data batch is the only owner, and then transitions to idle, it cannot unsubscribe (compile error). Worth adding a test for this case.

kevkrist (changes requested):

include/cucascade/data/data_batch.hpp
dhruv9vats: great catch, thanks @kevkrist! fixed in [make (un)subscribe const too](https://github.com/NVIDIA/cuCascade/pull/192/commits/73406bc8412b4876bec4f52d46b1b0564053ac0b).

kevkrist (approved):
[fistbump]

9prady9 (approved):

PR #196, opened 2026-09-15
fix: Use libcucascade domain for all NVTX ranges

9prady9:
## Summary

- move every cuCascade NVTX range into the `libcucascade` domain
- replace the converter's global-domain `nvtxRangePushA`/`nvtxRangePop` wrapper with `nvtx3::scoped_range_in`
- remove the `CUCASCADE_NVTX` CMake option and always compile the annotations
- propagate the header-only `CUDA::nvtx3` target through the library's public usage requirements

## Rationale

NVTX dispatches to no-op stubs when no profiling tool is injected, so a separate compile-time switch is unnecessary. Always compiling the ranges ensures they are available whenever cuCascade is profiled. Keeping both the function-level and named `hg:*` ranges in `libcucascade` also makes cuCascade activity easy to distinguish from NVTX ranges emitted by other libraries.

## Verification

- configured the release build with CUDA 13.2 and cuDF 26.08
- compiled and linked `cucascade_shared` and `cucascade_cudf_shared` with warnings treated as errors
- verified generated CMake export targets propagate `CUDA::nvtx3`
- passed all pre-commit hooks on the changed files


mbrobbel:
/ok to test [edfbb63](https://github.com/NVIDIA/cuCascade/pull/196/commits/edfbb63fc46f858bc0d565868ceb31d83350adf3)

felipeblazing (approved):

mbrobbel (approved):

